PR #1, opened 2017-01-27
Added a basic Place model and a test that shows off GIS functionality.

leonardr:
This branch kicks off the library registry project by copying in the bare minimum code from the `core` module necessary to get a configuration file, a database schema, and a testing framework.

I define one data model class, `Place`, and write a test suite that shows that we can create GIS-aware database objects and run queries on them to do things like calculate distances.

### DIFF
--- a/model.py
+++ b/model.py
@@ -1,0 +1,176 @@
+from config import Configuration
+import logging
+import warnings
+from sqlalchemy import (
+    Column,
+    ForeignKey,
+    Integer,
+    String
+)
+from sqlalchemy import (
+    create_engine
+)
+from sqlalchemy.exc import (
+    IntegrityError
+)
+from sqlalchemy.ext.declarative import (
+    declarative_base
+)
+from sqlalchemy.orm import (
+    backref,
+    relationship,
+    sessionmaker,
+)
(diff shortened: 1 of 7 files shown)


model.py
```diff
@@ -0,0 +1,176 @@
...
+    kwargs.update(create_method_kwargs or {})
+    created = getattr(model, create_method, model)(**kwargs)
+    db.add(created)
+    db.flush()
+    return created, True
+
+    
+Base = declarative_base()
+
+class Place(Base):
```
leonardr: Except for this class and its test, all the code in this branch is a slightly modifed version of the same code in `core`.

model.py
```diff
@@ -0,0 +1,176 @@
...
+    NoResultFound,
+    MultipleResultsFound,
+)
+
+from geoalchemy2 import Geography
+
+def production_session():
+    url = Configuration.database_url()
+    if url.startswith('"'):
+        url = url[1:]
```
aslagle: I know this came from core but I'm curious what it's meant to handle.

leonardr: It was a workaround for a bug, like, two years ago. I've taken the code out of here and everything still seems to work.

PR #2, opened 2017-01-31
Load Place objects from GeoJSON data

leonardr:
This branch adds a script that takes the output of a script like geojson-places-us and loads the data into the database as Place objects.

The input of the script looks like:
```
[metadata about a place]
[GeoJSON object]
[metadata about a place]
[GeoJSON object]
...
```

The GeoJSON objects are on separate lines so I don't have to load the JSON, extract the GeoJSON, and convert it back to to a string to pass it into the ST_GeomFromGeoJSON function.

I added the `abbreviated_name` field to the `Place` model so it would have a place to go. I haven't added aliases yet because I want to make sure we're prepared for place names in different languages, but I figure the official abbreviations for places are language-independent.


geography_loader.py
```diff
@@ -0,0 +1,67 @@
...
+        if parent_external_id:
+            parent = self.places_by_external_id[parent_external_id]
+        else:
+            parent = None
+
+        # This gives us a Geometry object.
+        geometry = func.ST_GeomFromGeoJSON(geography)
+
+        # We want real-world coordinates. Convert it into a Geography.
+        geography = cast(geometry, Geography)
```
leonardr: Unfortunately there is no function to convert GeoJSON into a Geography object--we have to convert it to a Geometry and then cast it.

scripts.py
```diff
@@ -0,0 +1,96 @@
...
+import os
+import sys
+
+from geography_loader import GeographyLoader
+from model import (
+    production_session
+)
+from config import Configuration
+
+class Script(object):
```
leonardr: This is copied and slightly modified from core.

geography_loader.py
```diff
@@ -0,0 +1,66 @@
...
+        metadata = json.loads(metadata)
+        external_id = metadata['id']
+        type = metadata['type']
+        parent_external_id = metadata['parent_id']
+        name = metadata['name']
+        aliases = metadata.get('aliases', [])
+        abbreviated_name = metadata.get('abbreviated_name', None)
+
+        if parent_external_id:
+            parent = self.places_by_external_id[parent_external_id]
```
aslagle: Do the parents always come before their children in the input?

leonardr: Yes, I do the US first, then all the states, and everything else (counties, cities, ZIP) is underneath the states.

PR #3, opened 2017-02-01
Add the PlaceAlias table

leonardr:
This branch completes the import of data from geojson-places-us by adding a PlaceAlias class which contains alternate names for places.

I don't expect to be using PlaceAlias.language very soon, but I want it in place from the beginning. I think it will be useful for libraries in countries that have more than one official language, like Canada and South Africa.

The other big change here is I added `Place.geo`, which casts the `Place.geography` field to `Geography`. 
Our ORM defines the field as a Geography field, but internally the data is stored as though it were a Geometry field. Without the cast, this means database functions treat places as shapes on a flat plane of undetermined dimensions, rather than shapes on the Earth with distances measured in meters.

I was not able to find any way of forcing the `geography` field to always be treated as Geometry without an expicit cast. Supposedly "Standard geometry type data will autocast to geography if it is of SRID 4326." That is the case for this field but it's not happening...

### DIFF
--- a/geography_loader.py
+++ b/geography_loader.py
@@ -7,6 +7,7 @@ from sqlalchemy.sql.expression import cast
 from model import (
     get_one_or_create,
     Place,
+    PlaceAlias,
 )
 
 class GeographyLoader(object):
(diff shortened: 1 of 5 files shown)


leonardr:
The Geography only becomes a Geometry when it's written to the database and looked up again. Until I introduced PlaceAliases, this never happened during the tests. I think because the ORM objects were cached.

PR #5, opened 2017-02-01
Get from a place to a list of libraries that serve that place

leonardr:
This branch adds `Place.served_by`, which goes from a known Place (such as a ZIP code, city or state) to a list of libraries whose service areas intersect that place. This will drive all the searches that aren't based on current location or the name of the library.

There's one big problem to do with borders, which I deal with in a way I think is a little hacky. I think there's a way around this and will be talking to Bert Spaan about it tomorrow.


model.py
```diff
@@ -266,6 +267,31 @@ def geo(self):
...
+            # We don't do this for cities because it's much more
+            # likely that a library will accept patrons from the next
+            # town over.
+            #
+            # TODO: With ST_ContainsProperly we might be able to
+            # eliminate this extra code, but that function doesn't
+            # work on geometry objects.
+            qu = qu.filter(
+                or_(ServiceArea.place==self, Place.parent==self)
+            )
```
aslagle: Would this work for counties? Or would cities always have the state as their parent even if there's also a county?

leonardr: Cities, ZIP codes and counties all have the state as their parent.

PR #6, opened 2017-02-01
Create `LibraryAlias` and implement basic search by library name

leonardr:
This branch creates the concept of "library alias", which is almost exactly the same as "place alias", but for libraries. It's used to create alternate names for libraries (e.g. "NYPL" for the New York Public Library).

This branch also implements `Library.by_name`, which performs a slightly fuzzy match against a library's official name, and a less fuzzy match against its aliases.

### DIFF
--- a/model.py
+++ b/model.py
@@ -161,8 +161,13 @@ class Library(Base):
     @classmethod
     def for_name(cls, _db, name):
         """Find a library whose name or alias matches the given name."""
+
+        # We allow for minor misspellings in the official name,
+        # but not in aliases (which are likely to be acronyms)
+        name_close_enough = func.levenshtein(func.lower(Library.name),
+                                             func.lower(name)) < 2
         qu = _db.query(Library).outerjoin(Library.aliases).filter(
-            or_(Library.name.ilike(name), LibraryAlias.name.ilike(name))
+            or_(name_close_enough, LibraryAlias.name.ilike(name))
         )
         return qu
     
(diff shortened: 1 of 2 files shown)


model.py
```diff
@@ -155,9 +155,23 @@ class Library(Base):
...
+    def for_name(cls, _db, name):
+        """Find a library whose name or alias matches the given name."""
+
+        # We allow for minor misspellings in the official name,
+        # but not in aliases (which are likely to be acronyms)
+        name_close_enough = func.levenshtein(func.lower(Library.name),
+                                             func.lower(name)) < 2
+        qu = _db.query(Library).outerjoin(Library.aliases).filter(
+            or_(name_close_enough, LibraryAlias.name.ilike(name))
+        )
```
aslagle: This looks good but I don't think it's quite enough.

If we have NYPL's official name as "The New York Public Library", a search for "new york public library" or "new york public" won't find it because the distance is too high. "New York Public Library" can be an alias, but then a search for "new york public libary" still won't work if aliases need exact matches.

I think allowing misspellings in the alias if the length is more than 6-ish could work well.

aslagle: Distinguishing aliases that are abbreviations could also work.

leonardr: This is true, but I'm reluctant to put a lot of work into this until we know more about how people run searches. It might be worth it to create an Elasticsearch index just for library names.

leonardr: After investigating how people across the US search for their local library, we've come up with an overall plan for handling searches, which I'll implement in a future branch.

PR #7, opened 2017-02-02
Use geometry instead of geography

leonardr:
This branch switches the database over so that shapes are stored as geometry objects instead of geography objects. Most of this is renaming but there are two points I don't want to get lost in the renaming:

Right now the big advantage of geography objects is that they support the ST_Touches function, which lets us eliminate `Place`s that intersect only because they border each other (and thus do not share library service). This lets me get rid of a hack in `Place.served_by`. Supposedly there are also performance benefits, and I'll be talking with Bert Spaan about that later today.

The most important difference between geometry and geography objects is you need to store the SRID with them. Geometries with different SRIDs can't be compared. The SRID for normal planet Earth stuff is 4326. I have a helper method that sets the SRID on incoming GeoJSON objects, and I changed some tests to manually set the SRID as part of creating point geometries.

It's also important to use ST_Distance_Sphere (which gives you a result in meters) when comparing two geometries, instead of ST_Distance (which gives you a result in degrees). When I call ST_DWithin, the only solution I could come up with was to cast the geometry objects to geography objects. That's why I'm skeptical about the performance improvements. But maybe Bert can help me figure out how to convert kilometers to degrees in this situation.


tests/test_geometry_loader.py
```diff
@@ -49,13 +49,11 @@ def test_load(self):
...
-        # and Texas. This verifies that the GeoJSON shapes are
-        # imported as real-world geographies, not abstract
-        # geometries.
-        distance_func = func.ST_Distance(new_york.geo, texas_zip.geo)
+        # and Texas.
+        distance_func = func.ST_Distance_Sphere(new_york.geometry, texas_zip.geometry)
         distance_qu = self._db.query().add_columns(distance_func)
         [[distance]] = distance_qu.all()
-        eq_(2511, int(distance/1000))
+        eq_(2510, int(distance/1000))
```
leonardr: These numbers are slightly different because ST_Distance on a geography takes the shape of the Earth into account, whereas ST_Distance_Sphere assumes the Earth is a sphere. The difference doesn't matter for our purposes.

testing.py
```diff
@@ -113,24 +114,24 @@ def _library(self, name=None, service_areas=[]):
         return library                  
 
     def _place(self, external_id=None, external_name=None, type=None,
-               abbreviated_name=None, parent=None, geography=None):
-        if not geography:
-            geography = 'POINT(%s %s)' % (
+               abbreviated_name=None, parent=None, geometry=None):
+        if not geometry:
+            geometry = 'SSRID=4326;POINT(%s %s)' % (
```
aslagle: there's an extra 'S' here

leonardr: I guess I was thinking of SSRIs.

PR #9, opened 2017-02-03
Implement geographic and name-based library search

leonardr:
This branch adds a search algorithm that should get us to 'better than Overdrive' status. A search returns two sets of results:

1. libraries with a name or alias that match your search query
2. libraries whose service area intersects a place whose name or alias matches your search query

The one likely case I can think of that this doesn't cover is a query like "city st" or "city, state".


model.py
```diff
@@ -103,6 +111,19 @@ def get_one(db, model, on_multiple='error', **kwargs):
     except NoResultFound:
         return None
 
+def dump_query(query):
```
leonardr: Copied from core.

model.py
```diff
@@ -218,6 +226,190 @@ def nearby(cls, _db, latitude, longitude, max_radius=150):
...
+        # We tack on any additional libraries that match a place query.
+        if place_query:
+            libraries_for_location = cls.search_by_location_name(
+                _db, place_query, place_type, here,
+                libraries_for_name
+            ).limit(max_libraries).all()
+        else:
+            libraries_for_location = []
+            
+        return libraries_for_name + libraries_for_location
```
aslagle: If a library matches both queries, it should only appear once in the final list.

aslagle: oh I see, you're excluding them in search_by_location_name

aslagle: If it doesn't affect performance, I think it would be clearer to do it here.

model.py
```diff
@@ -218,6 +226,190 @@ def nearby(cls, _db, latitude, longitude, max_radius=150):
...
+
+    @classmethod
+    def search_by_library_name(cls, _db, name, here=None):
+        """Find libraries whose name or alias matches the given name.
+
+        :param name: Name of the library to search for.
+        :param here: Order results by proximity to this location.
```
leonardr: This is a good idea but like you say I don't think it will make a big difference. Let's keep it in mind.

tests/test_model.py
```diff
@@ -198,3 +170,176 @@ def test_nearby(self):
...
+        manhattan_ks = self.manhattan_ks
+        
+        # A search for 'manhattan' finds both libraries.
+        libraries = list(Library.search_by_location_name(self._db, "manhattan"))
+        eq_(set(["NYPL", "Kansas State Library"]),
+            set([x.name for x in libraries])
+        )
+
+        # If you're searching from California, the Kansas library
+        # shows up first.
```
aslagle: I don't think we should do anything about this now, but for this particular example, it would be better for NYPL to show up first.

leonardr: You'd need to triangulate between population and distance. No matter how big NYC gets, if you search for "Manhattan" in Kansas you should get the one in Kansas.

We'll find out when people start using it, but I think a lot of the improvements beyond this point will not be cost-effective. Unlike book search, this is an interface people will probably only use once every few years.

PR #11, opened 2017-02-06
Add initial OPDS support

leonardr:
This branch adds some fields to the `libraries` table which we want to use in OPDS feeds, and implements the ability to turn a list of libraries into an OPDS navigation feed. The OPDS generation code is copied from  LS core. I also copied the idea of an `Annotator`, which will be used when I start writing the app server to insert links to the search interface, and any other app-server-specific information.

### DIFF
--- a/opds.py
+++ b/opds.py
@@ -154,7 +154,7 @@ class NavigationFeed(OPDSFeed):
     @classmethod
     def library_entry(cls, library):
         entry = AtomFeed.entry(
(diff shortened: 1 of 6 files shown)


opds.py
```diff
@@ -0,0 +1,195 @@
+import datetime
```
leonardr: The first part of this file is copied from core.

PR #12, opened 2017-02-06
Add some command-line scripts

leonardr:
This branch creates three useful command-line scripts:

1. `search_places` lets you search for places by name and find their `external_id`s.
2. `add_library` lets you create a `Library` object and associate a number of `external_id`s with its service area.
3. `search` is a command-line interface to the library search algorithm.

With these three scripts along with `load_places`, I was able to verify that the search algorithm performs reasonably well in a dev environment that includes one library and every place in the United States. I created a library representing NYPL and was able to quickly check whether or not a query ("10018", "Kew Gardens", "Brooklyn", "Kings", "New York") matched a place name inside NYPL's service area.

### DIFF
--- a/tests/test_scripts.py
+++ b/tests/test_scripts.py
@@ -5,10 +5,13 @@ from nose.tools import (
 from StringIO import StringIO
 
 from model import (
+    Library,
     Place,
 )
 from scripts import (
+    AddLibraryScript,
     LoadPlacesScript,
+    SearchPlacesScript,
 )
 from . import (
     DatabaseTest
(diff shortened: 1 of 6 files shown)


tests/test_scripts.py
```diff
@@ -34,3 +38,75 @@ def test_run(self):
...
+        eq_("Serving the five boroughs of New York, NY.", library.description)
+
+        [alias] = library.aliases
+        eq_("NYPL", alias.name)
+        eq_("eng", alias.language)
+
+        eq_([nyc], [x.place for x in library.service_areas])
+
+
+class TestSearchPlacesScript(DatabaseTest):
```
aslagle: this class should be TestSearchLibraryScript

PR #13, opened 2017-02-07
Implemented web app and distance in search results

leonardr:
This branch copies in a lot of code from `content` and `core` to run a basic web app. The actual implementation of the web app is new, however. We have a heartbeat controller, a 'libraries near me' controller, and a search controller.

I started incorporating the distance from the users' current location into the OPDS feed, using the schema:distance tag.


model.py
```diff
@@ -339,7 +345,7 @@ def search_by_location_name(cls, _db, query, type=None, here=None):
         # For a library to match, the Place named by the query must
         # intersect a Place served by that library.
         named_place = aliased(Place)
-        qu = _db.query(Library).join(
+        qu = _db.query(Library).distinct().join(
```
leonardr: Without this distinct() call, a library will show up more than once it it serves two different areas, and those areas are different distances away from `here`. This wasn't a problem when we weren't selecting `distance`.

util/app_server.py
```diff
@@ -0,0 +1,125 @@
+"""Implement logic common to more than one of the Simplified applications."""
```
leonardr: All the stuff in util/ is copied from core and only modified slightly, if at all.

util/app_server.py
```diff
@@ -0,0 +1,125 @@
...
+from nose.tools import set_trace
+from psycopg2 import DatabaseError
+import flask
+import json
+import sys
+from lxml import etree
```
aslagle: lxml needs to be added to requirements.txt

controller.py
```diff
@@ -0,0 +1,131 @@
...
+            return None
+        return GeometryUtility.point_from_ip(ip_address)
+        
+    def nearby(self, ip_address):
+        point = self.point_from_ip(ip_address)
+        qu = Library.nearby(self._db, point)
+        qu = qu.limit(5)
+        this_url = self.app.url_for('nearby')
+        feed = NavigationFeed(
+            self._db, "Find your library", this_url, qu,
```
aslagle: I think the title of this feed should tell you that it's showing you nearby libraries.

model.py
aslagle: Will it use the shorter distance for that library?

opds.py
```diff
@@ -148,16 +148,23 @@ def __init__(self, _db, title, url, libraries, annotator=None):
...
         entry = AtomFeed.entry(
             AtomFeed.id(library.urn_uri),
             AtomFeed.title(library.name),
             AtomFeed.updated(cls._strftime(library.timestamp))
         )
+
+        if distance is not None:
+            distance_tag = AtomFeed.SCHEMA.distance()
+            distance_tag.text = "%d km." % (distance/1000)
+            entry.append(distance_tag)
```
aslagle: Are you thinking the app would show this to patrons?

controller.py
aslagle: this title could also be localized since you brought in all the localization code.

controller.py
```diff
@@ -0,0 +1,131 @@
...
+        
+    def search(self, ip_address=None):
+        point = self.point_from_ip(ip_address)
+        query = flask.request.args.get('q')
+        if query:
+            # Run the query and send the results.
+            results = Library.search(self._db, point, query)
+            this_url = self.app.url_for('search', q=query)
+            feed = NavigationFeed(
+                self._db, 'Search results for "%s"' % query,
```
aslagle: this could also be localized

opds.py
leonardr: Yeah, it's a possibility. It doesn't really cost anything to include this information, and making it visible would explain why the results are in the order they are.

controller.py
leonardr: I meant to ask you about this. I tried localizing these strings but I ran into a problem generating the OPDS feed. The line `self.E.title(title),` caused an error:

```
TypeError: bad argument type: LazyString(<flask_babel.speaklater.LazyString object at 0x7fbb3e47cc90>)
```

I went looking in the code and couldn't find a place where we localize the title of an OPDS feed. Do you have any ideas? I think we probably need to address this as a separate task.

aslagle: You probably need to convert the LazyString to a string at the point where it's actually used - `self.E.title(str(title))` or `self.E.title(unicode(title))`

aslagle: The code that converts it needs to run as part of the request, so it will have the accept language header.

aslagle: Since this is a controller, you could even do it right here. LazyString is more for constants that are defined outside a request, but I think it's easier to use the same localization function everywhere even if some of the uses don't need to be lazy.

model.py
leonardr: Probably? I think it will find the shorter distance first, because of the query ordering and then ignore the library when it shows up again with a longer distance. But even if that's wrong, the distances will be comparable in most cases, because they'll be adjacent cities or different parts of a single city.

controller.py
leonardr: That seems to work, thanks!

PR #14, opened 2017-02-07
Start importing (some) Adobe Vendor ID code

leonardr:
This branch starts the process of making the library registry an Adobe Vendor ID server. 

For now, all I have done is port the code that encodes and decodes short client tokens (the things used to look up Adobe IDs). Encoding will primarily happen on the circulation manager and the encoding code is only here for testing purposes. Decoding will happen only in the library registry, as part of the implementation of the Adobe Vendor ID server. This will let me eventually remove code from the circulation manager.

The algorithm is mostly the same as in the circulation manager, with one big difference. In the circulation manager, information about libraries (their short names, their secrets, etc.) is kept in the Configuration object. In the library registry, we have Library objects, and Adobe-specific information about those libraries is kept in the database. The only thing that needs to be kept in the library registry's configuration is the node value we got from Adobe.

Basically, I am less concerned here than in the circ manager about creating an absolute separation between the data model and the Adobe code. The circulation manager has a very complicated data model and the Adobe code has been getting more and more complicated as we try more and more things to work around Adobe's brokenness. Here we're starting fresh and because this is already a piece of software that tracks lots of different libraries, adding the Adobe functionality means adding only one database table.


model.py
```diff
@@ -601,3 +624,167 @@ class PlaceAlias(Base):
...
+        identifier, is_new = get_one_or_create(
+            _db, DelegatedPatronIdentifier, library=library,
+            patron_identifier=patron_identifier, type=identifier_type
+        )
+        if is_new:
+            identifier.delegated_identifier = create_function()
+        return identifier, is_new
+
+
+class ShortClientTokenDecoder(ShortClientTokenTool):
```
aslagle: why is this class in model.py?

leonardr: This class decodes a short client token into a DelegatedPatronIdentifier. (The circ manager version decodes it into two strings, `library_uri` and `patron_identifier`) I put it in model because I think of it as a helper class for DelegatedPatronIdentifier.

PR #15, opened 2017-02-08
Copy adobe vendor ID code

leonardr:
This branch copies all the remaining Adobe vendor ID code that can be copied in from circulation without modification. This includes the code that parses incoming requests, handles the abstract logic of the Vendor ID spec, and generates responses.

It does not include any code that interacts with the data model or the web application -- that code needs to be ported to the new model/application and simplified, and I'll handle it in the next branch.

One bit of code has been modified: the code that gathers Configuration values for the vendor ID implementation. There's almost no code now that most of the configuration information is kept in `Library` objects; all we need to keep track of now are two scalar values: the vendor ID itself and the node value.

### DIFF
--- a/config.py
+++ b/config.py
@@ -132,7 +132,7 @@ class Configuration(object):
             return None, None
         return (
             integration[cls.ADOBE_VENDOR_ID],
-            integration[cls.ADOBE_NODE_VALUE],
+            integration[cls.ADOBE_VENDOR_ID_NODE_VALUE],
         )
     
     @classmethod
--- a/config.py
+++ b/config.py
@@ -1,6 +1,22 @@
+import contextlib
+import copy
 import json
 import os
 import logging
+
+@contextlib.contextmanager
+def temp_config(new_config=None, replacement_classes=None):
+    old_config = Configuration.instance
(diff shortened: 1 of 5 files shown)


config.py
```diff
@@ -101,6 +121,21 @@ def database_url(cls, test=False):
         return cls.integration(cls.DATABASE_INTEGRATION)[key]
 
     @classmethod
+    def vendor_id(cls):
```
leonardr: This is the only new code in this branch.

tests/test_adobe_vendor_id.py
```diff
@@ -0,0 +1,201 @@
...
+        """Configure a basic Vendor ID Service setup."""
+        name = Configuration.ADOBE_VENDOR_ID_INTEGRATION
+        with temp_config() as config:
+            config[Configuration.INTEGRATIONS][name] = {
+                Configuration.ADOBE_VENDOR_ID: "VENDORID",
+                Configuration.ADOBE_VENDOR_ID_NODE_VALUE: 114740953091845,
+            }
+            yield config
+
+class TestConfiguration(VendorIDTest):
```
leonardr: This is the test for the only new code in this branch.

PR #16, opened 2017-02-08
Reimplement Adobe Vendor ID

leonardr:
This branch implements the portions of our crusty Vendor ID implementation that still make sense.

The only authentication mechanism supported is authentication via short client token. You can send a short client token as an authdata, or you can split it into username and password. In either case, if the signature matches the shared secret we have on file for the originating `Library`, a DelegatedPatronIdentifier is created for you, and your Adobe Account ID is the UUID associated with that DelegatedPatronIdentifier.

I connected the vendor ID controller to the web app, and added a script that I've been using to simulate the Adobe acceptance test. Using this script I was able to validate real short client tokens from NYPL and Brooklyn.

I changed the create-a-library script so that you can specify the Adobe short name and shared secret for that library.

Here's the stuff that's in the circ manager implementation but _not_ in this branch:

There is no way to create a persistent authdata token by authenticating with library credentials, because the library registry doesn't check library credentials.

There is no way to authenticate with a persistent authdata token, either as authdata or by smuggling it in as a username, because there is no way to create a persistent authdata token.

There is no way to get your vendor ID with your library credentials, because the library registry doesn't check library credentials.

There is no support for authenticating with an authdata JWT. The JWTs are useless because short client tokens do the same job and Adobe won't support using authdata to deactivate devices.


adobe_vendor_id.py
```diff
@@ -143,3 +178,59 @@ def handle_accountinfo_request(self, data, urn_to_label):
...
+
+    """Implement Adobe Vendor ID within the library registry's database
+    model.
+    """
+
+    def __init__(self, _db, node_value, temporary_token_duration=None):
+        self._db = _db
+        self.temporary_token_duration = (
+            temporary_token_duration or datetime.timedelta(minutes=10)
+        )
```
aslagle: does the token duration need to be here? isn't that set on the circ manager side?

leonardr: No, the circ manager specifies the time the token was _issued_. The library registry decides whether it was issued recently enough to be acceptable.

aslagle: Then should the decoder be using this? It's using the time in the token as an expiration time now.

leonardr: Hmm, I must be confused one way or the other then.

leonardr: You're right, the circ manager specifies the time the token _expires._ I'll fix the registry.

PR #20, opened 2017-02-13
If a library's coverage area includes two places with the same name, we only want to select the library once.

leonardr:
Many ZIP codes have 'Brooklyn' as an alias, and NYPL has all those ZIP codes in its coverage area, so searching for 'brooklyn' picked up NYPL once for each ZIP code. This branch eliminates duplicates by only finding the _minimum_ distance between the library's coverage area and _any_ place that matches the search.


testing.py
```diff
@@ -201,7 +204,17 @@ def zip_10018(self):
...
+
+    @property
+    def zip_11212(self):
+        place = self._place(
+            '11212', '11212', Place.POSTAL_CODE, None, self.new_york_state,
+            self.zip_11212_geojson
+        )
+        alias = get_one_or_create(
+            self._db, PlaceAlias, place=place, name='Brooklyn'
+        )
```
aslagle: I guess it didn't matter for the test, but this method should return something

PR #22, opened 2017-02-14
Delegate unrecognized authentication attempts to another Vendor ID server

leonardr:
This branch makes it possible for a library registry to take any standard or authdata credentials it doesn't understand and ask another Vendor ID server to validate them. This gives us a migration path for moving from another Vendor ID server to the library registry.

### DIFF
--- a/adobe_vendor_id.py
+++ b/adobe_vendor_id.py
@@ -18,12 +18,17 @@ class AdobeVendorIDController(object):
     Authorization Service portions of the Adobe Vendor ID protocol.
     """
     def __init__(self, _db, vendor_id, node_value,
-                 external_credential_validator):
+                 delegates=[]):
+        """Constructor.
+        
+        :param delegates: A list of URLs or AdobeVendorIDClient
+        objects. If this Vendor ID server cannot validate an incoming
+        login, it will delegate to each of these other servers in
+        turn.
+        """
         self._db = _db
         self.request_handler = AdobeVendorIDRequestHandler(vendor_id)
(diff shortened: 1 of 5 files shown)


adobe_vendor_id.py
```diff
@@ -231,3 +277,141 @@ def account_id_and_label(self, delegated_patron_identifier):
...
+
+class VendorIDAuthenticationError(Exception):
+    """The Vendor ID service is working properly but returned an error."""
+
+    
+class VendorIDServerException(Exception):
+    """The Vendor ID service is not working properly."""
+    
+
+class AdobeVendorIDClient(object):
```
leonardr: This code is moved in from `AdobeVendorIDAcceptanceTestScript`, which now uses it. I don't have unit tests for it, but I have re-tested it against the library registry, and I have written unit tests for the way it's used in the controller above.

tests/test_adobe_vendor_id.py
```diff
@@ -285,3 +290,49 @@ def test_short_client_token_lookup_failure(self):
...
+        delegate2.enqueue(("userid", "label", "content"))
+
+        result = self.model.standard_lookup(
+            dict(username="some", password="user")
+        )
+        eq_(("userid", "label"), result)
+        
+        # We tried delegate 1 before getting the answer from delegate 2.
+        eq_([], delegate1.queue)
+        eq_([], delegate1.queue)
```
aslagle: should the second line here be delegate2?

leonardr: Yes.

Issue #24, opened 2017-07-12
Implement registry side of library sync


leonardr:
Once a library is configured, we need to get it into the library registry somehow. Thanks to [our extensions to Authentication For OPDS](https://github.com/NYPL-Simplified/Simplified/wiki/Authentication-For-OPDS-Extensions), all the necessary information will be stored in the library's Authentication For OPDS document. All we need is to be notified of the location of that document.

We need to define an endpoint on the library registry that accepts the URL to an OPDS server and an optional shared secret (for when a library wants to change its shared secret). 

The library registry hits the provided URL and locates an Authentication For OPDS document.

It imports information about the library from the Authentication For OPDS document.

If a library for the given URL already exists in this system, locally stored information is updated.

If the library has a public key, a shared secret is generated (if necessary) and encrypted with the library's public key. A short library name is generated or looked up.

If there is not enough information to perform registration, or any of the information is inconsistent with existing information, an error document is returned.

Otherwise success is returned. If the library provided a public key, the success message includes the encrypted shared secret and the short library name.

PR #27, opened 2017-07-21
Added controller for registering libraries.

aslagle:
This starts #24. The register endpoint accepts an OPDS URL, uses the shelf link to get the authentication document, and extracts library information from that.


testing.py
```diff
@@ -233,3 +234,39 @@ def zip_11212(self):
...
+        response = self.responses.pop()
+
+        code = response.status_code
+        series = "%sxx" % (code / 100)
+
+        if allowed_response_codes and (
+            code not in allowed_response_codes and series not in allowed_response_codes
+            ):
+            raise BadResponseException(url, "Bad Response!", status_code=code)
+        return response
```
aslagle: This is brought over from core, but I changed it so I could use it in place of HTTP.get_with_timeout instead of having to define another method that calls that.

util/http.py
```diff
@@ -0,0 +1,256 @@
+from nose.tools import set_trace
```
aslagle: This entire file is from core.

tests/test_util_http.py
```diff
@@ -0,0 +1,286 @@
+import requests
```
aslagle: This file is copied from core.

controller.py
```diff
@@ -78,6 +84,10 @@ def annotate_feed(self, feed):
         feed.add_link_to_feed(
             feed.feed, href=search_url, rel="search", type=OPENSEARCH_MEDIA_TYPE
         )
+        register_url = self.app.url_for("register")
+        feed.add_link_to_feed(
+            feed.feed, href=register_url, rel="register"
+        )
```
leonardr: I assume we're using the semantics of the 'register' link relation defined in the Authentication for OPDS spec.

aslagle: Yes, except that we're registering a library on the registry instead of registering a user account on an OPDS server.

aslagle: Well, maybe not quite since you can also use it to update an existing library. Should we use a different rel?

controller.py
```diff
@@ -138,3 +148,50 @@ def search(self, ip_address=None):
...
+            response = do_get(opds_url, allowed_response_codes=["2xx", "3xx"])
+            feed = feedparser.parse(response.content)
+        except Exception, e:
+            return INVALID_OPDS_FEED
+        links = feed.get("feed", {}).get("links", [])
+
+        auth_url = None
+        for link in links:
+            # TODO: Define a rel for the auth document itself, and
+            # look for that instead.
```
leonardr: The link relation exists, it's `http://opds-spec.org/auth/document`. I suspect Hadrien added it to the A4OPDS spec recently, because I remember planning to define it and being surprised that it was already in there.

Forcing a 401 by getting the shelf is a good fallback if there is no http://opds-spec.org/auth/document. So here's what I'd do:

1. GET the registration URL.
2. If it's a 401, read the auth document from the body
3. Otherwise, GET the auth document.
4. If no link to auth document, GET the shelf link and hope for a 401.
5. Give up and complain.

controller.py
```diff
@@ -138,3 +148,50 @@ def search(self, ip_address=None):
...
+            auth_document = json.loads(response.content)
+        except Exception, e:
+            return INVALID_AUTH_DOCUMENT
+
+        library, is_new = get_one_or_create(
+            self._db, Library,
+            opds_url=opds_url
+        )
+
+        library.name = auth_document.get("name")
```
leonardr: It might be premature, but given the number of extensions we've added to A4OPDS I think it's a good idea to create a data object that can parse an A4OPDS document. If nothing else it'll make it a lot easier to write tests that handle various weird documents.

problem_details.py
```diff
@@ -0,0 +1,30 @@
...
+      _("Invalid OPDS feed"),
+      _("The submitted URL did not return a valid OPDS feed."),
+)
+
+NO_SHELF_LINK = pd(
+    "http://librarysimplified.org/terms/problem/no-shelf-link",
+    400,
+    title=_("No shelf link"),
+    detail=_("The submitted OPDS feed did not have a link with rel 'http://opds-spec.org/shelf'.")
+)
```
leonardr: Change this to a generic "you have an OPDS server but I couldn't find an auth document" error.

testing.py
leonardr: Smart idea. :mortar_board:

controller.py
leonardr: Let's go with register for now. I've asked Hadrien and Mike Amundsen if they know of something else we can use.

aslagle:
@leonardr I made those changes. 200 response code from shelf link (!), link to auth document in body is weird, but I handled that case anyway.

controller.py
```diff
@@ -154,27 +154,69 @@ def register(self, do_get=HTTP.get_with_timeout):
...
+                        auth_response = response
+                    else:
+                        # This response didn't require authentication, so maybe it's a feed
+                        # that links to the auth document.
+                        feed = feedparser.parse(response.content)
+                        links = feed.get("feed", {}).get("links", [])
+                        for link in links:
+                            if link.get("rel") == "http://opds-spec.org/auth/document":
+                                auth_response = do_get(link.get("href"), allowed_response_codes=["2xx", "3xx"])
+                                break
```
leonardr: I think this + line 177 can be refactored.

leonardr:
:+1:

PR #29, opened 2017-07-25
Place parser

leonardr:
This branch creates the model code necessary for parsing the coverage areas described in an Authentication for OPDS document. Strings like "Lake County, OH", "Buffalo, NY" and "New York State" can be parsed and turned into the appropriate Place objects.

There are some places we should be able to disambiguate but can't, and some places that can be identified in weird ways, but we can handle all the place names defined in the A4OPDS spec, and I don't think there's a case where we identify the _wrong_ place.


model.py
```diff
@@ -585,11 +575,167 @@ class Place(Base):
...
+        There is no guarantee that these place names correspond to
+        Places in the database.
+
+        :param name: The name to split into parts.
+        :return: A list of place names, with the largest place at the front
+           of the list.
+        """
+        return [x.strip() for x in reversed(name.split(",")) if x.strip()]
+    
+    def strictly_intersects(self, qu):
```
aslagle: I would call this "strictly_inside". "strictly_intersects" sounds like the opposite of this to me.

aslagle: Although, this can be either "strictly inside" or "strictly contains", so maybe "strictly inside" isn't right either.

PR #30, opened 2017-07-25
Converted controllers to return OPDS 2 catalogs.

aslagle:
This branch changes the library registry to use [OPDS 2](https://github.com/opds-community/opds-revision/blob/master/opds-2.0.md), with a custom "catalogs" collection role. Within "catalogs", the metadata and navigation collection are the same as in the overall OPDS 2 spec.

The register endpoint also returns an OPDS catalog now, but it doesn't have a self link yet.

### DIFF
--- a/opds.py
+++ b/opds.py
@@ -28,16 +28,12 @@ class OPDSCatalog(object):
     @classmethod
     def add_link_to_catalog(cls, catalog, children=None, **kwargs):
         link = dict(**kwargs)
-        links = catalog.get("links", [])
-        links.append(link)
-        catalog["links"] = links
+        catalog.setdefault("links", []).append(link)
 
     @classmethod
     def add_image_to_catalog(cls, catalog, children=None, **kwargs):
         image = dict(**kwargs)
-        images = catalog.get("images", [])
-        images.append(image)
-        catalog["images"] = images
+        catalog.setdefault("images", []).append(image)
 
     def __init__(self, _db, title, url, libraries, annotator=None):
         """Turn a list of libraries into a catalog."""
--- a/opds.py
+++ b/opds.py
@@ -7,6 +7,12 @@ class Annotator(object):
         pass
 
 class OPDSCatalog(object):
+    """Represents an OPDS 2 Catalog document.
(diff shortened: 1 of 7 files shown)


opds.py
```diff
@@ -1,202 +1,97 @@
...
+        return date.strftime(cls.TIME_FORMAT)
 
     @classmethod
-    def name(cls, *args, **kwargs):
-        return cls.E.name(*args, **kwargs)
+    def add_link_to_catalog(cls, catalog, children=None, **kwargs):
+        link = dict(**kwargs)
+        links = catalog.get("links", [])
+        links.append(link)
+        catalog["links"] = links
```
leonardr: You can save some code here with `catalog.setdefault("links", []).append(link)`

opds.py
```diff
@@ -1,202 +1,97 @@
...
 
     @classmethod
-    def summary(cls, *args, **kwargs):
-        return cls.E.summary(*args, **kwargs)
+    def add_image_to_catalog(cls, catalog, children=None, **kwargs):
+        image = dict(**kwargs)
+        images = catalog.get("images", [])
+        images.append(image)
+        catalog["images"] = images
```
leonardr: Similarly here.

opds.py
```diff
@@ -1,202 +1,97 @@
...
-
+class Annotator(object):
 
-    @classmethod
-    def entry(cls, *args, **kwargs):
-        return cls.E.entry(*args, **kwargs)
+    def annotate_feed(self, feed):
+        pass
 
+class OPDSCatalog(object):
```
leonardr: I was confused for a while between the concept of the "OPDS Catalog" and the "catalogs" role _within_ one of these catalogs which links to _other_ catalogs. A docstring here saying that this represents an OPDS 2 Catalog document would have helped, I think.

aslagle: I was thinking the catalogs in the "catalogs" role would also be "OPDS Catalogs" themselves.

leonardr: I think that's the right thing to do, but the recursion is another thing that confused me. Saying that this object might be an OPDS document on its own, or it might be contained within another OPDS document using the "catalogs" role, would help.

PR #31, opened 2017-07-26
Auth document parser

leonardr:
This branch introduces the `AuthenticationDocument` class, which represents the result of parsing an Authentication For OPDS document (with our extensions). JSON coverage objects are parsed into Place objects, links are extracted, etc.

I noticed two problems when looking at the Authentication for OPDS documents we currently serve. First, we send the name of the library as `name` rather than `title`. Second, we don't send the `type` field. Instead, we send `providers`, which is either an old name for what is now called `type` or is something slightly different--at any rate, I think it's not compliant with the current spec. I'll investigate this more later.

### DIFF
--- a/authentication_document.py
+++ b/authentication_document.py
@@ -1,0 +1,178 @@
+import json
+
+class AuthenticationDocument(object):
+    """Parse an Authentication For OPDS document, including the
+    Library Simplified-specific extensions, extracting all the information
+    that's of interest to the library registry.
+    """
+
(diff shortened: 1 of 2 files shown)


aslagle:
@leonardr you're right about `providers` not being compliant with the spec. The spec doesn't support multiple providers so I had to come up with something else, and it didn't get added to the spec. I also never documented it :(

aslagle:
However, I don't think that information is needed by the registry so it shouldn't affect this pr.

leonardr:
Merging while I sort out the `types` thing with Hadrien.

PR #33, opened 2017-07-28
Use auth document

aslagle:
This branch converts the registration controller method to parse the OPDS auth document with AuthenticationDocument, and adds support for registering service areas and logo links.

AuthenticationDocument.parse_coverage wasn't working with the Place class, so I had to refactor it a bit and make MockPlace more like Place.

### DIFF
--- a/controller.py
+++ b/controller.py
@@ -258,7 +258,12 @@ class LibraryRegistryController(object):
         if auth_document.service_area:
             places, unknown, ambiguous = auth_document.service_area
             if unknown or ambiguous:
-                return INVALID_AUTH_DOCUMENT.detailed(_("The authentication document has an unknown or ambiguous service area."))
+                msgs = []
+                if unknown:
+                    msgs.append(str(_("The following service area was unknown: %(service_area)s.", service_area=json.dumps(unknown))))
+                if ambiguous:
+                    msgs.append(str(_("The following service area was ambiguous: %(service_area)s.", service_area=json.dumps(ambiguous))))
+                return INVALID_AUTH_DOCUMENT.detailed(" ".join(msgs))
             for place in places:
                 service_area = get_one_or_create(self._db, ServiceArea,
                                                  library_id=library.id,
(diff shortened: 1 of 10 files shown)


model.py
```diff
@@ -482,14 +482,6 @@ def urn_uri(self):
         else:
             return 'urn:' + self.urn
     
-    @property
-    def logo_data_uri(self):
-        """Return the logo as a data: URI."""
-        if not self.logo:
-            return None
-        return "data:image/png;base64,%s" % base64.b64encode(self.logo)
```
aslagle: This isn't necessary - the logo is stored as a data URI in the db.

model.py
```diff
@@ -577,7 +569,7 @@ class Place(Base):
     # The geography of the place itself. It is stored internally as a
     # geometry, which means we have to cast to Geography when doing
     # calculations.
-    geometry = Column(Geometry(srid=4326))
+    geometry = Column(Geometry(srid=4326), nullable=True)
```
aslagle: I was getting an error from Place.everywhere because it had a null geometry. I'm not sure why it didn't break tests.

controller.py
```diff
@@ -225,19 +229,49 @@ def find_and_get_url(links, rel, allowed_response_codes=None):
...
+            type = logo_response.headers.get("Content-Type") or auth_document.logo_link.get("type")
+            if type:
+                library.logo = "data:%s;base64,%s" % (type, b64)
+        else:
+            library.logo = None
+
+        if auth_document.service_area:
+            places, unknown, ambiguous = auth_document.service_area
+            if unknown or ambiguous:
+                return INVALID_AUTH_DOCUMENT.detailed(_("The authentication document has an unknown or ambiguous service area."))
```
leonardr: It would be useful to specify which bits were unknown or ambiguous, especially because "unknown" might mean we just don't know about a place that really exists (as opposed to, you made a typo).

controller.py
```diff
@@ -225,19 +229,49 @@ def find_and_get_url(links, rel, allowed_response_codes=None):
...
+            for place in places:
+                service_area = get_one_or_create(self._db, ServiceArea,
+                                                 library_id=library.id,
+                                                 place_id=place.id)
+            for service_area in library.service_areas:
+                if service_area.place_id not in [p.id for p in places]:
```
leonardr: Build this list while you're iterating over `places` above, or at least outside the `for` loop so you don't create a nested loop.

model.py
leonardr: I also find this weird because I thought I set nullable=True in one of my branches.

PR #34, opened 2017-07-31
Bring opds registration up to spec

aslagle:


controller.py
```diff
@@ -168,6 +170,22 @@ def register(self, do_get=HTTP.get_with_timeout):
         SHELF_REL = "http://opds-spec.org/shelf"
 
         auth_response = None
+        def get_opds_links(response):
+            type = response.headers.get("Content-Type")
+            if type == "application/opds+json":
```
leonardr: Why no startswith here?

aslagle: OPDS 2 doesn't have different kinds of feeds like OPDS 1, so it's not necessary. I guess it wouldn't hurt to have it anyway.

leonardr: It does, because of `profile`, but right now the only profile is the directory one, and we don't want to import another directory.

aslagle: In that case maybe it shouldn't have startswith.

leonardr: Yeah, not for now, certainly.

Issue #25, opened 2017-07-12
Add flag that hides libraries until they're ready to go live


leonardr:
Libraries need to be tested before being introduced into the main feed. But they can't be tested if they're not in the library registry at all. We need a flag that can be set to push a library "live".

We also need a separate feed that ignores the flag, so that we can use SimplyE to test libraries that aren't ready to go yet.

leonardr:
While designing the circulation manager-metadata wrangler interface I decided that we need to keep track of _three_ states:

1. Library has been registered but no one has looked at it.
2. Library has been approved. It represents a bona fide library and we've verified the server is actually run by that organization.
3. Library is live, and ready to show up in SimplyE, to the delight of all.

PR #36, opened 2017-08-01
Create a 'qa' feed (as distinct from the 'live' feed) containing only items in the 'approved' stage.

leonardr:
This branch introduces an alternate feed at /qa which shows only items in the "approved" stage. (The default feed shows only items in the "live" stage.) This is done with alternate inputs into the `nearby` and `search` controllers.

### DIFF
--- a/app.py
+++ b/app.py
@@ -55,6 +55,13 @@ def nearby():
         originating_ip()
     )
 
+@app.route('/qa')
+@returns_problem_detail
+def nearby_qa():
+    return app.library_registry.registry_controller.nearby(
+        originating_ip(), qa=True
+    )
+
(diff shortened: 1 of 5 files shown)


leonardr:
In support of https://github.com/NYPL-Simplified/library_registry/issues/25

app.py
```diff
@@ -55,6 +55,13 @@ def nearby():
         originating_ip()
     )
 
+@app.route('/qa')
+@returns_problem_detail
+def nearby_qa():
+    return app.library_registry.registry_controller.nearby(
+        originating_ip(), qa=True
```
aslagle: The controllers have a 'live' kwarg but this uses 'qa'.

PR #38, opened 2017-08-02
Introduce different types of ServiceArea, such as 'focus area'

leonardr:
This branch introduces the concept of 'focus area' by adding a `type` field to ServiceArea. The type of ServiceArea we've been dealing with up to this point has `type="eligibility"`, and the new type has `type="focus"`.

A Library advertises its focus area by naming specific places in its Authentication For OPDS document.  I refactored some code from the controller into `AuthenticationDocument` and made it work both for `service_area` and `focus_area`.

Any ServiceArea indicates that people in a certain Place are served by the Library, but the "focus area" indicates a Place that is the Library's primary concern. By default, a Library's focus area is the same as its service area.

We should be prepared for a library to have the same Place as two different ServiceAreas with different types, but in the most common case where this would happen (the library has a service area but no focus area) we avoid this by only registering the focus area.

### DIFF
--- a/authentication_document.py
+++ b/authentication_document.py
@@ -116,7 +116,7 @@ class AuthenticationDocument(object):
             # parse anything.
             place_objs.append(place_class.everywhere(_db))
             coverage = dict()
-            
+
         for country, places in coverage.items():
             try:
                 country_obj = place_class.lookup_one_by_name(
@@ -286,9 +286,9 @@ class AuthenticationDocument(object):
                 msgs.append(str(_("The following service area was unknown: %(service_area)s.", service_area=json.dumps(unknown))))
             if ambiguous:
                 msgs.append(str(_("The following service area was ambiguous: %(service_area)s.", service_area=json.dumps(ambiguous))))
+            _db.rollback()
             return INVALID_AUTH_DOCUMENT.detailed(" ".join(msgs))
 
-        place_ids = []
         for place in places:
             service_area, is_new = get_one_or_create(
                 _db, ServiceArea, library_id=library.id,
--- a/authentication_document.py
+++ b/authentication_document.py
@@ -1,14 +1,21 @@
 from collections import defaultdict
 import json
 from nose.tools import set_trace
+from flask.ext.babel import lazy_gettext as _
 from sqlalchemy.orm.exc import (
     MultipleResultsFound,
     NoResultFound,
 )
 
 from model import (
-    Place
+    get_one_or_create,
+    Place,
+    ServiceArea,
 )
+
+from problem_details import INVALID_AUTH_DOCUMENT
+from sqlalchemy.orm.session import Session
+
 
 class AuthenticationDocument(object):
     """Parse an Authentication For OPDS document, including the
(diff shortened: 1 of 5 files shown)


authentication_document.py
```diff
@@ -45,13 +52,13 @@ def __init__(self, _db, id, title, type, service_description, color_scheme,
                 _db, service_area, place_class=place_class
             )
         else:
-            self.service_area = [place_class.everywhere(_db)], [], []
+            self.service_area = [place_class.everywhere(_db)], {}, {}
```
leonardr: parse_coverage returns a list and two dicts, and this should use the same data types.

authentication_document.py
```diff
@@ -45,13 +52,13 @@ def __init__(self, _db, id, title, type, service_description, color_scheme,
...
         if focus_area:
             self.focus_area = self.parse_coverage(
                 _db, focus_area, place_class=place_class
             )
         else:
-            self.focus_area = [place_class.everywhere(_db)], [], []
+            self.focus_area = self.service_area
```
leonardr: Default focus area is the service area, not 'everywhere'.

leonardr:
Part of https://github.com/NYPL-Simplified/library_registry/issues/39

authentication_document.py
```diff
@@ -203,7 +210,88 @@ def from_dict(cls, _db, data, place_class=Place):
...
+        places, unknown, ambiguous = areas
+        if unknown or ambiguous:
+            msgs = []
+            if unknown:
+                msgs.append(str(_("The following service area was unknown: %(service_area)s.", service_area=json.dumps(unknown))))
+            if ambiguous:
+                msgs.append(str(_("The following service area was ambiguous: %(service_area)s.", service_area=json.dumps(ambiguous))))
+            return INVALID_AUTH_DOCUMENT.detailed(" ".join(msgs))
+
+        place_ids = []
```
aslagle: you don't use this array.

controller.py
```diff
@@ -316,24 +316,9 @@ def find_and_get_url(links, rel, allowed_response_codes=None):
...
-                service_area = get_one_or_create(self._db, ServiceArea,
-                                                 library_id=library.id,
-                                                 place_id=place.id)
-                place_ids.append(place.id)
-            for service_area in library.service_areas:
-                if service_area.place_id not in place_ids:
-                    self._db.delete(service_area)
+        problem = auth_document.update_service_areas(library)
+        if problem:
+            return problem
```
aslagle: You may need to rollback the db in this case - I had some cases in the circ manager where returning a problem detail didn't roll back since it wasn't raising an exception. And you might have created a library, or updated an eligibility area when there was a subsequent problem with the focus area.

model.py
```diff
@@ -241,7 +241,8 @@ class Library(Base):
         "DelegatedPatronIdentifier", backref='library'
     )
     service_areas = relationship('ServiceArea', backref='library')
-
+    focus_areas = relationship('ServiceArea', backref='library_focus')
```
aslagle: how does this work?

controller.py
leonardr: OK, I've added a rollback. I had to move some of the controller tests to separate functions because rolling back undid the setup step.

model.py
leonardr: I wrote that immediately before realizing that `type` was a better solution.

Issue #39, opened 2017-08-02
Gather more information from Authentication For OPDS document


leonardr:
We need to gather and store the following information:

* Focus area (in addition to service area)
* Audiences
* Collection size
* Is anonymous access allowed?
* Can you register for a card online?

PR #40, opened 2017-08-03
Add audiences

leonardr:
This branch adds an `audiences` table which has a many-to-many relationship with `libraries`. A library's audiences come in through the Authentication For OPDS document. By default, a library's audience is the general public.


leonardr:
I'm not sure about "everybody plus you equals everybody". My goal was to stop libraries from checking every single audience box representing all the people they _hope_ will use the library. But there are libraries like IA that have subcollections for the general public and subcollections that mainly appeal to researchers. I would want IA to show up in a general search _and_ in a research-specific search. So maybe I need to get over my aggravation that libraries in general will want to check all the boxes.

leonardr:
In service of https://github.com/NYPL-Simplified/library_registry/issues/39

authentication_document.py
```diff
@@ -215,6 +216,60 @@ def from_dict(cls, _db, data, place_class=Place):
...
+    def _update_audiences(self, library, audiences):
+        original_audiences = audiences
+        if not audiences:
+            audiences = [Audience.PUBLIC]
+        if isinstance(audiences, basestring):
+            # This is invalid but we can easily support it.
+            audiences = [audiences]
+        if not isinstance(audiences, list):
+            return INVALID_AUTH_DOCUMENT.detailed(
+                _("'audience' must be a list") % audiences
```
aslagle: String formatting doesn't work with flask-babel. You need to do `_("'audience' must be a list: %(audiences)r", audiences=audiences)`. Same for the other problem detail below.

authentication_document.py
```diff
@@ -215,6 +216,60 @@ def from_dict(cls, _db, data, place_class=Place):
...
+        # whole document.
+        audiences = [x for x in audiences
+                     if x in Audience.KNOWN_AUDIENCES]
+
+        # But there must be at least one audience we recognize.
+        if original_audiences and not audiences:
+            return INVALID_AUTH_DOCUMENT.detailed(
+                _("None of the provided audiences were recognized: %r") %
+                original_audiences
+            )
```
aslagle: You could use Audience.OTHER here - what's it for if not this case?

model.py
```diff
@@ -833,6 +832,47 @@ class PlaceAlias(Base):
...
+    id = Column(Integer, primary_key=True)
+    name = Column(Unicode, index=True, unique=True)
+
+    libraries = relationship("Library", secondary='libraries_audiences',
+                             back_populates="audiences")
+    
+    @classmethod
+    def lookup(cls, _db, name, allow_unknown=False):
+        if not allow_unknown and name not in cls.KNOWN_AUDIENCES:
+            raise ValueError("Unknown audience: %s" % anme)
```
aslagle: you spelled name wrong

authentication_document.py
leonardr: It's for use by a library that wants to say they appeal to some other audience. Using it here is a good idea; my only concern is that if the library uses OTHER they're supposed to spell out who the audience is in their service_description, and if they think they've spelled it out in their `audience` they probably won't bother.

aslagle:
For advanced search, I do think it would be useful for libraries to specify general public and another audience. Someone might want to filter to libraries for researchers or for people with print disabilities, and you wouldn't want to exclude libraries that have materials for the general public as well.

PR #41, opened 2017-08-03
Collection size

leonardr:
Captures https://github.com/NYPL-Simplified/Simplified/wiki/Authentication-For-OPDS-Extensions#collection-size in service of https://github.com/NYPL-Simplified/library_registry/issues/39


tests/test_languages.py
```diff
@@ -0,0 +1,67 @@
+# encoding: utf-8
```
leonardr: This file was copied from core.

tests/test_language.py
```diff
@@ -0,0 +1,67 @@
+# encoding: utf-8
```
leonardr: The tests of language.py were missing, so I copied them from core.

leonardr:
I can't figure out why a seemingly unrelated test is failing. I'm still working on it; see if anything jumps out at you.

leonardr:
I think SQLAlchemy didn't like mixing the technique we used to update Library.service_areas (delete the IDS that don't fit) with the technique we used to update Library.collections (just set the property to a new list).

model.py
```diff
@@ -872,8 +876,50 @@ def lookup(cls, _db, name):
...
+    language = Column(Unicode)
+    size = Column(Integer)
+
+    @classmethod
+    def set(cls, library, language, size):
+        """Create or update a CollectionSummary for the given
+        library and language.
+
+        :return: An up-to-date CollectionSummary, or None if the language
+         is not recognized.
```
aslagle: This isn't true, it still returns a CollectionSummary if the language isn't recognized.

PR #42, opened 2017-08-03
Store booleans for 'anonymous access' and 'online registration'

leonardr:
This branch closes out https://github.com/NYPL-Simplified/library_registry/issues/39 by storing the last two pieces of information we need from the authentication document: whether anonymous access is allowed and whether you can get library credentials through an entirely online process.

### DIFF
--- a/controller.py
+++ b/controller.py
@@ -284,9 +284,6 @@ class LibraryRegistryController(object):
             create_method_kwargs=dict(stage=Library.REGISTERED)
         )
 
-        library.name = auth_document.title
-        library.description = auth_document.service_description
-
         if auth_document.website:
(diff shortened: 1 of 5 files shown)


controller.py
```diff
@@ -284,9 +284,6 @@ def find_and_get_url(links, rel, allowed_response_codes=None):
             create_method_kwargs=dict(stage=Library.REGISTERED)
         )
 
-        library.name = auth_document.title
-        library.description = auth_document.service_description
-
```
leonardr: I'd like to move a lot more code out of here and into `AuthenticationDocument.update_librar()`, but the base URL and the possibly mocked `do_get` make it a little tricky, so I didn't do much.

PR #44, opened 2017-08-04
Expect New OPDS auth format

leonardr:
This branch changes the library registry to handle the near-final Authentication for OPDS format. The main changes:

* Rather than a list of authentication types and a single set of authentication configuration, an auth document has a list of authentication objects, each of which has a `type` as well as its own configuration.
* An authentication document has a list of links, with `rel` as an attribute within each link, rather than an object with `rel` values keyed to sub-lists of links. An authentication flow can have its own list of links, which is separate from the main list.

### DIFF
--- a/authentication_document.py
+++ b/authentication_document.py
@@ -208,7 +208,8 @@ class AuthenticationDocument(object):
             return None
         good_enough = None
         if not isinstance(links, list):
-            raise ValueError("links must be a list (got %r)" % links)
+            # Invalid links object; ignore it.
+            return
         for link in links:
             if rel != link.get('rel'):
                 continue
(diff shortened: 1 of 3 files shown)


authentication_document.py
```diff
@@ -65,7 +66,7 @@ def __init__(self, _db, id, title, type, service_description, color_scheme,
         self.website = self.extract_link(
             rel="alternate", require_type="text/html"
         )
-        self.online_registration = self.extract_link(rel="register") is not None
+        self.online_registration = self.has_link(rel="register")
```
leonardr: This is a judgment call on my part; I figure if there is a single authentication mechanism that has a `register` link, we can treat the library as one that offers online registration.

PR #45, opened 2017-08-10
Added query to rank libraries by relevance.

aslagle:
This is an implementation of https://github.com/NYPL-Simplified/Simplified/wiki/LibraryRegistrySearchAlgorithm

This works reasonably well with the example scenarios we came up with, but will need to be tested and adjusted as we add more libraries to the system. 

I'm also not sure how the performance will be.


model.py
```diff
@@ -280,7 +289,217 @@ def _restrict_to_stages(cls, qu, allowed_stages=None):
...
+            libraries_collections.c.collectionsummaries_id,
+            libraries_collections.c.collectionsummaries_size,
+        ).order_by(
+            score.desc()
+        )
+
+        result = _db.execute(library_id_and_score)
+        library_ids_and_scores = [r for r in result]
+        # Look up the Library objects and return them with the scores.
+        return [(get_one(_db, Library, id=lid), score) for lid, score in library_ids_and_scores]
```
aslagle: It would be better if the original query returned the Library objects, but I couldn't figure out a way to do that, since I had to use the SQL expression API instead of ORM.

model.py
```diff
@@ -280,7 +289,216 @@ def _restrict_to_stages(cls, qu, allowed_stages=None):
...
+        :param target: The user's current location. May be a Geometry object or
+        a 2-tuple (latitude, longitude).
+        :param language: The ISO 639-1 code for the user's language.
+        :param audiences: List of audiences the user is a member of.
+        By default, only libraries with the PUBLIC audience are shown.
+        :param allowed_stages: A list of allowable values for
+        `Library.stage`. By default, only libraries in the LIVE stage
+        are shown.
+
+        :return A list of (Library, score) tuples, sorted by relevance.
```
aslagle: I'm not sure what this should return. The scores are useful when testing it, but probably not necessary anywhere else.

leonardr: I think leave it like this. We can put the score in the feed and use it to troubleshoot problems.

model.py
leonardr: How about getting all the libraries with a single query against Library.id and then organizing them into a Counter based on the scores? That will save a lot of database hits.

PR #46, opened 2017-08-25
Compare SCTs with low-numbered expiration dates against a new epoch.

leonardr:
This fixes https://github.com/NYPL-Simplified/circulation/issues/648 by shaving several characters off the 'username' portion of the short client token.

### DIFF
--- a/tests/test_short_client_token.py
+++ b/tests/test_short_client_token.py
@@ -176,7 +176,7 @@ class TestShortClientTokenDecoder(DatabaseTest):
 
         # The patron identifier must not be blank.
         assert_raises_regexp(
-            ValueError, 'Token library|1234| has empty patron identifier',
+            ValueError, 'Token library\|1234\| has empty patron identifier',
             m, self._db, "library|1234|", "signature"
         )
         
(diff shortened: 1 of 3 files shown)


tests/test_short_client_token.py
```diff
@@ -176,7 +176,7 @@ def test_short_client_token_lookup_delegated_patron_identifier_failure(self):
 
         # The patron identifier must not be blank.
         assert_raises_regexp(
-            ValueError, 'Token library|1234| has empty patron identifier',
+            ValueError, 'Token library\|1234\| has empty patron identifier',
```
leonardr: The pipe characters were being used when this string was compiling into a regular expression, causing some tests to pass when they shouldn't have.

util/short_client_token.py
```diff
@@ -26,13 +26,26 @@ def adobe_base64_decode(cls, str):
...
         return base64.decodestring(encoded)
 
-    EPOCH = datetime.datetime(1970, 1, 1)
+    # The JWT spec takes January 1 1970 as the epoch.
+    JWT_EPOCH = datetime.datetime(1970, 1, 1)
+
+    # For the sake of shortening tokens, the Short Client Token spec
+    # takes January 1 2017 as the epoch, and measures time in minutes
+    # rather than seconds.
+    SCT_EPOCH = datetime.datetime(2017, 1, 1)
```
leonardr: Apart from these constants, none of the code in this file makes a difference, but I changed it to be consistent with what the circ manager will be generating in the future.

PR #48, opened 2017-09-05
Fix registration

leonardr:
This branch updates the library registry to use the new version of the [OPDS Directory Registration Protocol](https://github.com/NYPL-Simplified/Simplified/wiki/OPDS-Directory-Registration-Protocol), incorporating changes I made to get a real registration working between registry and circulation manager.

The biggest change is that the registration URL must be the URL to the Authentication For OPDS document, not the URL to the OPDS feed. This lets us get rid of a lot of complex logic. However, I had to add the additional requirement that a compatible Authentication for OPDS document must link to the root of the OPDS feed using the `start` link relation.


testing.py
```diff
@@ -325,6 +325,12 @@ def queue_response(self, response_code, media_type="text/html",
     def do_get(self, url, headers=None, allowed_response_codes=None, **kwargs):
         self.requests.append(url)
         response = self.responses.pop()
+        if isinstance(response.status_code, Exception):
+            raise response.status_code
```
leonardr: This lets you simulate an exception (such as a timeout) that happened during the request.

PR #49, opened 2017-09-05
Crosscheck opds feed

leonardr:
This branch adds a crosscheck to the registration process. Previously, the Authentication For OPDS document had to have a `start` link, but we didn't do anything to check on the document at the other end of this link. Now we verify that a) this document is an OPDS 1 or OPDS 2 document, and b) it links back to the Authentication For OPDS document that was used to initialize registration in the first place.

This prevents Alice from setting up a fake Authentication For OPDS document for Bob's site and gaining control of Bob's registry entry.

This is similar to code we used to have, but we go auth document -> OPDS feed instead of OPDS feed -> (maybe another OPDS feed, or maybe the first OPDS feed was actually an auth document) -> auth document. The main new thing that's been added is support for the `Link` tag and relative links.

One case we don't support is the case where getting the OPDS feed returns a 401 with an auth document in the body.


controller.py
```diff
@@ -192,41 +193,96 @@ def search(self, ip_address=None, live=True):
...
-                auth_url, exc_info=e
-            )
-            return AUTH_DOCUMENT_TIMEOUT
-        except Exception, e:
-            logging.error(
-                "Registration of %s failed: error retrieving authentication document", 
-                exc_info=e
-            )
-            return ERROR_RETRIEVING_AUTH_DOCUMENT
+        def _make_request(url, on_404, on_timeout, on_exception):
```
leonardr: This method is used to get the A4OPDS document and the OPDS feed that it links to. If there's a problem, a generic problem detail document is detailed with more specific errors.

testing.py
```diff
@@ -313,14 +315,14 @@ def __init__(self):
...
-                       other_headers=None, content=''):
+                       other_headers=None, content='', links=None):
         headers = {}
         if media_type:
             headers["Content-Type"] = media_type
         if other_headers:
             for k, v in other_headers.items():
                 headers[k.lower()] = v
-        self.responses.append(DummyHTTPResponse(response_code, headers, content))
+        self.responses.insert(0, DummyHTTPResponse(response_code, headers, content, links))
```
leonardr: This makes `queue_response` act like a queue instead of a stack, which is more accurate to the name and which makes the tests easier to read.

aslagle:
So if your feed requires authentication, you're required to include the Link header when returning a 401 with the auth document?

leonardr:
I looked at the code and the 401 case isn't supported at all. I'm adding support for it now.

problem_details.py
```diff
@@ -8,32 +8,38 @@
...
-    title=_("Authentication document not found"),
-    detail=_("No OPDS authentication document was present at the specified URL."),
+    title=_("Document not found"),
+)
+
+AUTH_DOCUMENT_TIMEOUT = pd(
+      "http://librarysimplified.org/terms/problem/timeout",
+      408,
+      _("Request timed out"),
+      _("Attempt to retrieve the OPDS root feed timed out."),
```
aslagle: I don't think you're using this problem detail (and if you are, its name doesn't match its description).

leonardr:
OK, take another look.

PR #51, opened 2017-09-08
Script that sets a library's coverage area

leonardr:
This branch introduces a script for setting a library's service area and focus area. We can use this script to set this information on a per-library basis until we develop a service area editor as part of the circulation manager admin interface.

I refactored a lot of the area-parsing code in `authentication_document.py` so I could call it outside the context of a full authentication document, and moved `MockPlace` into `testing.py` since it's now needed in multiple testing modules.


aslagle:
Do you have a script file in bin/ you forgot to add?

PR #52, opened 2017-09-27
Add logging

leonardr:
This branch removes the old (hardly used) logging code and replaces it with a copy of the logging code I added to the `core` module yesterday. Pretty much all the code in this is directly taken from `core`.

### DIFF
--- a/log.py
+++ b/log.py
@@ -1,0 +1,229 @@
+from nose.tools import set_trace
+import datetime
+import logging
+import json
+import os
+import socket
+from config import Configuration
+from StringIO import StringIO
+from loggly.handlers import HTTPSHandler as LogglyHandler
+
+if not Configuration.instance:
+    Configuration.load()
+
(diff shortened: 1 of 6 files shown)


log.py
```diff
@@ -0,0 +1,229 @@
+from nose.tools import set_trace
```
leonardr: This entire file copied from core.

testing.py
```diff
@@ -131,6 +135,45 @@ def _library(self, name=None, eligibility_areas=[], focus_areas=[], audiences=No
         library.stage = stage
         return library
 
+    def _external_integration(self, protocol, goal=None, settings=None,
```
leonardr: Entire method copied from core.

tests/test_log.py
```diff
@@ -0,0 +1,172 @@
+import logging
```
leonardr: Entire file copied from core.

PR #53, opened 2017-09-28
Various tweaks to get logging to work properly.

leonardr:


log.py
```diff
@@ -98,7 +98,8 @@ def initialize(cls, _db, testing=False):
         # log level, which is probably higher than the normal log level.
         for logger in (
                 'sqlalchemy.engine', 'elasticsearch', 
-                'requests.packages.urllib3.connectionpool'
+                'requests.packages.urllib3.connectionpool',
+                'urllib3.connectionpool',
```
leonardr: Without this, the fact that something was logged to loggly will itself get logged to loggly, and so on in an infinite loop. We should make this change in core as well.

model.py
```diff
@@ -1364,6 +1364,17 @@ class ExternalIntegration(Base):
...
+    # put it here.
+    URL = u"url"
+
+    # If access requires authentication, these settings represent the
+    # username/password or key/secret combination necessary to
+    # authenticate. If there's a secret but no key, it's stored in
+    # 'password'.
+    USERNAME = u"username"
+    PASSWORD = u"password"
+
```
leonardr: All this code is taken from core. The tests didn't catch it because setting `ExternalIntegration.url` works fine in Python without this code, and the test passes -- the data just isn't stored in the database.

log.py
aslagle: Wouldn't that mean it should be excluded from logging entirely, not just set to a different level?

leonardr: I want to keep open the possibility that you might need to set urllib3.connectionpool log level to DEBUG on a temporary basis to diagnose a problem with an API. But any such API would probably be using requests.packages.urllib3.connectionpool, so I think it's okay to always set urllib3.connectionpool log level to WARN. Let me know what you think.

aslagle: yeah, I think that would be okay. Is this relying on a level threshold for sending to loggly?

leonardr: No, the loggly handler will send over everything it gets, which I think is the right thing to do. We just don't want it to trigger additional logs that it has to send over.

PR #56, opened 2018-07-02
Serve terms of service as part of a document that describes the registration process

leonardr:
This branch makes the registration endpoint for a library registry respond to GET as well as POST. A GET request will give an OPDS 2 catalog containing (at most) one link. This link is a `data:` link describing the terms of service for the registration endpoint. This is as opposed to the catalog endpoints, which  don't have any special terms of service, because their users are random library patrons who don't even know they're using a separate service. But in theory, there could be a different TOS for the registry as a whole.

The circulation manager will retrieve the registration TOS and require a would-be registrar to agree to it before proceeding.

Since this is the first library registry branch in nearly a year, I had to make a number of changes to make sure the code worked and the Travis tests ran with recent dependencies.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -16,7 +16,7 @@ git:
 
 before_install:
   - pip install "setuptools>=18.5"
-  - sudo apt-get install -y postgresql-9.3-postgis-2.3 postgis
+  - sudo apt-get install -y postgresql-9.3-postgis-2.3
   - sleep 10
 
 install:
--- a/app.py
+++ b/app.py
@@ -67,7 +67,7 @@ def nearby_qa():
         originating_ip(), live=False
     )
 
-@app.route("/register", methods=["GET,POST"])
+@app.route("/register", methods=["GET","POST"])
 @returns_problem_detail
 def register():
     return app.library_registry.registry_controller.register()
--- a/app.py
+++ b/app.py
@@ -4,7 +4,7 @@ import sys
 import urlparse
 
 from flask import Flask, url_for, redirect, Response, request
-from flask.ext.babel import Babel
+from flask_babel import Babel
 from flask_sqlalchemy_session import flask_scoped_session
 
 from config import Configuration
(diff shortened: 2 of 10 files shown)


.travis.yml
```diff
@@ -16,7 +16,7 @@ git:
 
 before_install:
   - pip install "setuptools>=18.5"
-  - sudo apt-get install -y postgresql-9.3-postgis-2.3 postgis
+  - sudo apt-get install -y postgresql-9.3-postgis-2.3
```
leonardr: It looks like `postgis` was never necessary, and now installing it brings in Postgres 10, which runs instead of the Postgres 9 we need.

app.py
```diff
@@ -4,7 +4,7 @@
 import urlparse
 
 from flask import Flask, url_for, redirect, Response, request
-from flask.ext.babel import Babel
+from flask_babel import Babel
```
leonardr: We now import Babel through a different package.

model.py
```diff
@@ -1420,23 +1420,23 @@ def url(self):
         return self.setting(self.URL).value
 
     @url.setter
-    def set_url(self, new_url):
+    def url(self, new_url):
```
leonardr: Following a change in SQLAlchemy, the setter method for a hybrid property needs to have the same name as the getter method.

leonardr:
This addresses https://jira.nypl.org/browse/SIMPLY-991.

.travis.yml
aslagle: It's not necessary because `postgresql-9.3-postgis-2.3` already pulls in the right version?

app.py
```diff
@@ -67,7 +67,7 @@ def nearby_qa():
         originating_ip(), live=False
     )
 
-@app.route("/register", methods=["POST"])
+@app.route("/register", methods=["GET,POST"])
```
aslagle: shouldn't this be `["GET", "POST"]`?

.travis.yml
leonardr: Yes, as far as I can tell that's right.

PR #57, opened 2018-07-03
Reject registration requests that don't include the three required email addresses

leonardr:
This branch addresses https://jira.nypl.org/browse/SIMPLY-467. Registrations will only succeed if a `contact` email address is provided along with the registration request, and the library's Authentication For OPDS includes an email address for patron support and an email address for the designated copyright agent.

Currently the email addresses are not stored anywhere and nothing is done with them -- they just have to be present at the time the library is registered.

### DIFF
--- a/controller.py
+++ b/controller.py
@@ -271,6 +271,25 @@ class LibraryRegistryController(object):
         headers = { "Content-Type": OPDS_CATALOG_REGISTRATION_MEDIA_TYPE }
         return Response(document, status, headers=headers)
 
+    def _required_email_address(self, uri, title):
+        """`uri` must be a mailto: URI.
+
+        :return: Either an email address or a customized ProblemDetail.
+        """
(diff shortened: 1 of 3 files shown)


controller.py
```diff
@@ -280,6 +281,13 @@ def register(self, do_get=HTTP.get_with_timeout):
         if not auth_url:
             return NO_AUTH_URL
 
+        integration_contact_uri = flask.request.form.get("contact")
+        integration_contact_email = self._required_email_address(
+            integration_contact_uri, "No valid integration contact address"
```
aslagle: Unless you plan to handle this problem detail specially in the circ manager, this message should match the language used in the admin interface so it's easier for the admin to find the missing setting when they see this.

leonardr:
Thanks, I'll update the language in the next branch, since I've already refactored this code.

PR #58, opened 2018-07-03
Store incoming email addresses associated with a library

leonardr:
This branch grabs the email addresses that come in associated with a library (the patron support address, the copyright designated agent address, and the configuration contact address) and stores them in a new table called Hyperlink. This is a lot simpler than the Hyperlink table in server_core but it's the same idea -- it's a URL (href) that has a specific relationship (rel) to a library.

Although a library may have multiple links of a given type, we only store one Hyperlink of a given type for a given library. The library's links may change over time, but we try to keep using the same value as long as that is a possibility.

### DIFF
--- a/controller.py
+++ b/controller.py
@@ -1,4 +1,3 @@
-from collections import defaultdict
 from nose.tools import set_trace
 import logging
 import flask
@@ -24,6 +23,7 @@ from authentication_document import AuthenticationDocument
 from model import (
     production_session,
     ConfigurationSetting,
+    Hyperlink,
     Library,
     ServiceArea,
     get_one_or_create,
@@ -287,6 +287,9 @@ class LibraryRegistryController(object):
         )
         if isinstance(integration_contact_email, ProblemDetail):
             return integration_contact_email
+        hyperlinks_to_create = [
+            (Hyperlink.INTEGRATION_CONTACT_REL, [integration_contact_email])
+        ]
 
         def _make_request(url, on_404, on_timeout, on_exception, allow_401=False):
             allowed_codes = ["2xx", "3xx", 404]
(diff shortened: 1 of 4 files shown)


controller.py
```diff
@@ -283,10 +283,14 @@ def register(self, do_get=HTTP.get_with_timeout):
...
         integration_contact_email = self._required_email_address(
-            integration_contact_uri, "No valid integration contact address"
+            integration_contact_uri,
+            "Invalid or missing configuration contact email address"
         )
         if isinstance(integration_contact_email, ProblemDetail):
             return integration_contact_email
+        hyperlinks_to_create = [
+            (Hyperlink.INTEGRATION_CONTACT_REL, [integration_contact_email])
+        ]
```
leonardr: We build up a list of Hyperlink objects we need to create, and then create them by calling set_hyperlink once a Library object exists.

controller.py
```diff
@@ -359,23 +363,18 @@ def _make_request(url, on_404, on_timeout, on_exception, allow_401=False):
...
+        # patrons to get help or file a copyright complaint. Store these
+        # links in the database as Hyperlink objects.
+        links = auth_document.links or []
         for rel, problem_title in [
-                ('help', "No valid patron help email address"),
-                ("http://librarysimplified.org/rel/designated-agent/copyright",
-                 "No valid copyright designated agent email address")
+            ('help', "Invalid or missing patron support email address"),
+            (Hyperlink.COPYRIGHT_DESIGNATED_AGENT_REL,
+             "Invalid or missing copyright designated agent email")
```
leonardr: I changed the language here to match the language in the circulation manager. I need to make a few small changes to the circulation manager, which I'll put into my next circulation branch.

PR #59, opened 2018-07-05
Create a way to track the validation of an email address

leonardr:
This branch adds a data model class called Validation, used to keep track of the validation of an email address. Validating an email address requires presenting a secret key to a new controller. Eventually we'll send out emails containing links to that controller.

To avoid the need to validate the same email address over and over again (e.g. because a consortium with 30 libraries uses the same address for their patron support and their integration support), I split out Hyperlink into two classes: Hyperlink and Resource. This setup is more like what we have in the server core: Resource talks about the URI and Hyperlink is just a join table between the URI and the Library. Once the Resource is validated, it's validated for everyone.

The downside of this is that a rogue library can claim their patron support address is an address belonging to some other library that's already been validated. It'll be the responsibility of the registry administrator to deal with any such situations.

### DIFF
--- a/controller.py
+++ b/controller.py
@@ -21,12 +21,14 @@ from adobe_vendor_id import AdobeVendorIDController
 from authentication_document import AuthenticationDocument
 
 from model import (
-    production_session,
     ConfigurationSetting,
     Hyperlink,
     Library,
     ServiceArea,
+    Validation,
+    get_one,
     get_one_or_create,
+    production_session,
 )
 from config import (
     Configuration,
@@ -545,3 +547,48 @@ class LibraryRegistryController(object):
             return problem
 
         return candidates
+
+
+class ValidationController(object):
+    """Validates Resources based on validation codes.
+
+    The validation codes were sent out in emails to the addresses that
+    need to be validated, or otherwise communicated to someone who needs
+    to click on the link to this controller.
+    """
+
+    MESSAGE_TEMPLATE = "<html><head><title>%(message)s</title><body>%(message)s</body></html>"
+
+    def __init__(self, app):
+        self.app = app
+        self._db = self.app._db
+
+    def html_response(self, status_code, message):
+        """Return a human-readable message as a minimal HTML page.
+
+        This controller is used by human beings, so HTML is better
+        than Problem Detail Documents.
+        """
+        headers = {"Content-Type": "text/html"}
+        page = self.MESSAGE_TEMPLATE % dict(message=message)
+        return Response(page, status_code, headers=headers)
+
+    def validate(self, secret):
+        """Validate a secret, or don't.
+
+        :return: A Response containing a simple HTML document.
+        """
+        if not secret:
+            return self.html_response(404, _("No validation code provided"))
+        validation = get_one(self._db, Validation, secret=secret)
+        if not validation:
+            error = _("Validation code %r not found") % secret
+            return self.html_response(404, error)
+        if not validation.active:
+            error = _("Validation code %r has expired. Re-register to get another code.") % secret
+            return self.html_response(400, error)
+        validation.mark_as_successful()
+
+        resource = validation.resource
+        message = _("You successfully validated %s.") % resource.href
+        return self.html_response(200, message)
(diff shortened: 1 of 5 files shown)


model.py
```diff
@@ -1211,16 +1218,114 @@ class Hyperlink(Base):
...
+    def href(self):
+        if not self.resource:
+            return None
+        return self.resource.href
+
+    @href.setter
+    def href(self, url):
+        _db = Session.object_session(self)
+        resource, is_new = get_one_or_create(_db, Resource, href=url)
+        self.resource = resource
```
leonardr: This hybrid property let me keep most of the old code, which assumes you can read and write Hyperlink.href.

model.py
```diff
@@ -794,8 +801,8 @@ def set_hyperlink(self, rel, *hrefs):
         _db = Session.object_session(self)
         hyperlink, is_modified = get_one_or_create(
             _db, Hyperlink, library=self, rel=rel,
-            create_method_kwargs=dict(href=default_href)
         )
+
         if hyperlink.href not in hrefs:
             # This is a preexisting Hyperlink whose href doesn't
             # match any of the given hrefs.
```
aslagle: I think this comment isn't quite right now. The Hyperlink may have been newly created with no href and get one here.

model.py
```diff
@@ -1211,16 +1218,114 @@ class Hyperlink(Base):
...
+    href = Column(Unicode, nullable=False, index=True, unique=True)
+    hyperlinks = relationship("Hyperlink", backref="resource")
+
+    # Every Resource may have at most one Validation. There's no
+    # need to validate it separately for every relationship.
+    validation_id = Column(Integer, ForeignKey('validations.id'),
+                           index=True)
+    validation = relationship(
+        "Validation", backref=backref("resource", uselist=False)
+    )
```
aslagle: I think our convention so far has been to put the relationship on the model class that doesn't have the id column.

It may not make a difference, but that way you can still see that there's a relationship when you only look at the Validation class.

model.py
```diff
@@ -1211,16 +1218,114 @@ class Hyperlink(Base):
...
+    EXPIRES_AFTER = datetime.timedelta(days=1)
+
+    id = Column(Integer, primary_key=True)
+    success = Column(Boolean, index=True, default=False)
+    started_at = Column(DateTime, index=True, nullable=False,
+                        default = lambda x: datetime.datetime.utcnow())
+
+    # The only way to validate a Resource is to prove you know the
+    # corresponding secret.
+    secret = Column(Unicode, default=generate_secret)
```
aslagle: I'd make this unique just in case.

aslagle:
For security it would probably be good to send a confirmation email to the resource when a new library uses it, so they can contact the registry admin if it's a rogue library. It might be annoying if you're registering a lot of libraries, but if not it might be nice to get the confirmation.

controller.py
```diff
@@ -545,3 +547,48 @@ def _locate_email_addresses(cls, rel, links, problem_title):
...
+        if not secret:
+            return self.html_response(404, _("No validation code provided"))
+        validation = get_one(self._db, Validation, secret=secret)
+        if not validation:
+            error = _("Validation code %r not found") % secret
+            return self.html_response(404, error)
+        if not validation.active:
+            error = _("Validation code %r has expired. Re-register to get another code.") % secret
+            return self.html_response(400, error)
+        validation.mark_as_successful()
```
aslagle: This should handle the "You're already validated" case.

aslagle:
I think there's a potential problem if you register multiple libraries with the same emails before validating. Each registration will reset the secret so all but the last email will no longer work.

leonardr:
I changed the code to detect the "you're already validated case". The other issues you bring up I'll deal with when I hook up the registration process to the validation process. How about this:

I won't reset a validation on registration unless it's expired.

I _will_ send an email to an already validated address when a new library starts using it. If there is an active validation for that address, the email will mention the link to validate the address. Otherwise it will just be a notification.

This should take care of both cases you mention, at the expense of not being able to immediately reset a validation when you just didn't get the email -- worst case you'll have to wait a day.

PR #60, opened 2018-07-06
Switch over to looking in an environment variable for the database location.

leonardr:
This branch copies code from `core` so we can find the URL to the database in an environment variable (the preferred location) rather than in a JSON config file whose location is kept in an environment variable.

The only other item in the default JSON config file was the base URL, and the code wasn't even looking in the JSON config file for that information, so no more work is necessary to port the contents of the JSON config. While I was investigating this, I changed the web app startup to use a default URL in the case where this information was not in the database.

This is the component where a JSON config file would actually make sense, since there's no admin interface and only one instance, but it's such a pain to deploy modifications to the config file that I'd rather keep everything in the database.


config.py
```diff
@@ -43,20 +47,24 @@ class Configuration(object):
...
-        try:
-            cls.log.info("Loading configuration from %s", config_path)
-            configuration = cls._load(open(config_path).read())
-        except Exception, e:
-            raise CannotLoadConfiguration(
-                "Error loading configuration file %s: %s" % (
-                    config_path, e)
-            )
+        This is being phased out in favor of taking all configuration from a
+        database.
```
aslagle: How come we can't get rid of this now?

leonardr: I was just being conservative -- this code is still in the circ manager. I'd need to make sure we change over on the next deployment but since there's only one library registry and it hasn't been deployed in a year I'll need to take special care during the next deployment anyway.

PR #61, opened 2018-07-09
Implement a simple email system for sending validation emails

leonardr:
This branch introduces a very simple email templating system configured by a site-wide ExternalIntegration.

There are two kinds of emails: notification emails, which just inform you that someone associated your email address with a library in some capacity, and validation emails, which also include a link you can use to verify that your email address works.

The emails are sent through `Hyperlink.notify` -- they used to be set through `Validation.restart` but that doesn't make sense when you might want to send an email without resetting the validation.


emailer.py
```diff
@@ -0,0 +1,167 @@
...
+    PORT = 'port'
+    FROM_ADDRESS = 'from_address'
+
+    # Constants for different types of email.
+    VALIDATION = 'validation'
+    NOTIFICATION = 'notification'
+
+    EMAIL_TYPES = [VALIDATION, NOTIFICATION]
+
+    DEFAULT_NOTIFICATION_SUBJECT = "This address designated as the %(rel)s for %(library)s"
```
aslagle: I could imagine having other types of notifications in the future, e.g. your library has been approved, settings were updated...

emailer.py
```diff
@@ -0,0 +1,167 @@
...
+    DEFAULT_VALIDATION_SUBJECT = "Validate the %(rel)s for %(library)s"
```
aslagle: I think "Confirm" is more common language for these kind of emails

aslagle: and is the rel a human-readable string?

emailer.py
```diff
@@ -0,0 +1,167 @@
...
+        if not smtp_username:
+            raise CannotLoadConfiguration("No SMTP username specified")
+        self.smtp_username = smtp_username
+        if not smtp_password:
+            raise CannotLoadConfiguration("No SMTP password specified")
+        self.smtp_password = smtp_password
+        if not smtp_host:
+            raise CannotLoadConfiguration("No SMTP host specified")
+        self.smtp_host = smtp_host
+        if not smtp_host:
```
aslagle: this should be `smtp_port`

emailer.py
```diff
@@ -0,0 +1,167 @@
...
+    def body(self, from_address, to_address, **kwargs):
+        """Generate the complete body of the email message, including headers.
+
+        :param from_address: Originating address.
+        :param to_address: Destination address.
+        :param kwargs: Arguments to use when filling out the template.
+        """
+
+        message = email.Message.Message()
+        message['From'] = from_address
```
aslagle: should this be something like "Library Simplified Library Registry" instead of the email address?

emailer.py
```diff
@@ -0,0 +1,167 @@
...
+        if not smtp_host:
+            raise CannotLoadConfiguration("No SMTP host specified")
+        self.smtp_host = smtp_host
+        if not smtp_host:
+            raise CannotLoadConfiguration("No SMTP port specified")
+        self.smtp_port = smtp_port
+        if not from_address:
+            raise CannotLoadConfiguration("No From: address specified")
+        self.from_address = from_address
+        self.templates = templates
```
aslagle: It might be good to verify that no unexpected arguments are needed by the templates.

model.py
```diff
@@ -1224,6 +1232,8 @@ class Hyperlink(Base):
         UniqueConstraint('library_id', 'rel'),
     )
 
+    TIME_FORMAT = "%Y-%m-%d %H:%M:%S UTC"
```
aslagle: I don't think this is a very good choice for the time format. I actually might just say the link expires in X days instead of specifying the exact time.

leonardr: I took this out. The link expires one day from the time it was created, and since the email might only be checked once a day it's not really useful to talk in terms of days. In an earlier version of this branch the link could be sent out multiple times and this was more important.

emailer.py
leonardr: Yes, it is. I've renamed it to `rel_desc`.

leonardr:
This is ready for another look.

PR #63, opened 2018-07-11
Explain verification status of registered email addresses in a library's OPDS catalog

leonardr:
This branch addresses https://jira.nypl.org/browse/SIMPLY-468. The OPDS catalog for a library will include links to email addresses that have a confirmation step, with information about where they are in the confirmation process.

I use https://schema.org/reservationStatus and https://schema.org/ReservationStatusType to talk about the process, rather than making up my own vocabulary, since the email confirmation process is equivalent to the process of confirming a reservation.

This branch also removes (by commenting out) the requirement that the `contact` email address be provided on initial registration. Because of the strategy we've chosen to phase out Accounts.json, we need to support an interim period where circulation managers that haven't been upgraded to send `contact` can register with the library registry anyway.

### DIFF
--- a/controller.py
+++ b/controller.py
@@ -502,6 +502,9 @@ class LibraryRegistryController(object):
             encryptor = PKCS1_OAEP.new(public_key)
 
             if not library.short_name:
+                # TODO: This needs to be base64 encoded or something,
+                # not hex encoded. Hex encoding limits us to 256*3
+                # libraries. Also we need to handle collisions.
                 library.short_name = os.urandom(3).encode('hex')
 
             submitted_secret = None
--- a/controller.py
+++ b/controller.py
@@ -488,12 +488,14 @@ class LibraryRegistryController(object):
                 # them a new library is using their address.
                 hyperlink.notify(self.emailer, self.app.url_for)
                     
-        # Create an OPDS 2 catalog containing all publicly available
+        # Create an OPDS 2 catalog containing all available
         # information about the library.
-        catalog = OPDSCatalog.library_catalog(library)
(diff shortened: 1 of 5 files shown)


controller.py
```diff
@@ -469,16 +478,30 @@ def _make_request(url, on_404, on_timeout, on_exception, allow_401=False):
...
             return problem
+
+        for rel, candidates in hyperlinks_to_create:
+            hyperlink, is_modified = library.set_hyperlink(rel, *candidates)
+            if is_modified:
+                # We need to send an email to this email address about
+                # what just happened. This is either so the receipient
+                # can confirm that the address works, or to inform
+                # them a new library is using their address.
+                hyperlink.notify(self.emailer, self.app.url_for)
```
leonardr: Setting the hyperlinks before calling library_catalog ensures that the hyperlinks show up as links in the catalog.

controller.py
```diff
@@ -469,16 +478,30 @@ def _make_request(url, on_404, on_timeout, on_exception, allow_401=False):
...
 
+        # Annotate the catalog with some information specific to
+        # the transaction that's happening right now.
         public_key = auth_document.public_key
         if public_key and public_key.get("type") == "RSA":
             public_key = RSA.importKey(public_key.get("value"))
             encryptor = PKCS1_OAEP.new(public_key)
 
             if not library.short_name:
-                # TODO: Generate a short name based on the library's service area.
```
leonardr: This was a nice idea but I don't think it's worth the extra effort, since people commonly confuse this short name with the "short name" that's used in the circulation manager as the URL for designating a specific library.

PR #68, opened 2018-07-25
Treat library readiness for production as a two-part decision

leonardr:
This branch resolves https://jira.nypl.org/browse/SIMPLY-1030.
Previously a library had a single 'stage' which could be 'registered', 'approved', 'rejected', or 'live'. Now a library has a 'library_stage' (the library's opinion about which stage it should be in) and a 'registry_stage' (the registry's opinion on the same matter). Both stages can be 'testing', 'production', or 'cancelled'.

A library is in production if both the library and the registry think it should be. Once a library is in production, only the registry can take it out of production. (This is so random mistakes don't make libraries suddenly disappear from SimplyE.)

A library will show up in production feeds only if it is in production. A library will show up in QA feeds if neither the library or the registry has cancelled it.

I didn't write a migration script because before we launch I'll be rebuilding the libraries table completely, grandfathering in all libraries that are currently in the SimplyE system.

We had some leftover code that assumed the `id` in a library's Authentication for OPDS feed was an unchanging UUID, rather than the URL to the A4OPDS document itself. To get https://jira.nypl.org/browse/SIMPLY-1031 to work, we need to keep our own UUID that never changes, and use that as the library's `id` in the OPDS catalogs we serve. I made that change in this branch as long as I was changing the database schema.


tests/test_controller.py
```diff
@@ -164,21 +164,24 @@ def test_nearby(self):
             eq_("VENDORID", catalog["metadata"]["adobe_vendor_id"])
 
     def test_nearby_qa(self):
-        # The library we used in the previous test has stage=LIVE.
-        # If we switch to looking for libraries with stage=APPROVED,
-        # we won't find anything.
+        # The libraries we used in the previous test are in production.
+        # If we move them from production to TESTING, we won't find anything.
+        for library in self._db.query(Library):
+            library.registry_stage = Library.TESTING_STAGE
```
leonardr: In general, this code switched from "look for production libraries, then look for non-production libraries" to "look for production libraries, take them out of production, then verify they don't show up in the production feed, but they do show up in the QA feed."

model.py
```diff
@@ -298,30 +347,75 @@ def validate_short_name(self, key, value):
...
+
+    @library_stage.setter
+    def library_stage(self, value):
+        """A library can't unilaterally go from being in production to
+        not being in production.
+        """
+        if self.in_production and value != self.PRODUCTION_STAGE:
+            raise ValueError(
+                "This library is already in production; only the registry can take it out of production."
+            )
```
aslagle: What's the process for the registry to take it out of production? Bypass this setter and use `_library_stage`?

leonardr: The registry would set `registry_stage`, which doesn't require going through any code.

leonardr: Once registry_stage is no longer 'production', library_stage becomes unlocked and the library can set it to 'testing' or 'cancelled' if they want.

Since there's an out-of-band communication happening, the library can also leave library_stage alone and just coordinate with the library registry when the problem is fixed (assuming it's a problem that gets fixed).

leonardr: I guess what I'm saying is I don't think it's necessary to change `library_stage` to take a library out of production even if it was at the library's request.

aslagle: I could imagine a library wanting to be temporarily removed if they accidentally registered for production before they were ready, but it's probably unlikely, and you're right, they could coordinate with the registry admin.

PR #69, opened 2018-07-31
Create local DelegatedIdentifier objects for Adobe IDs obtained from delegate servers

leonardr:
Currently the AdobeVendorIDModel is responsible for keeping track of delegate servers. When it's asked to do a standard_lookup or an authdata_lookup, it asks the delegate servers first, and if they have an answer, it sends that answer to the client without doing anything else.

This leads to the race condition mentioned in https://jira.nypl.org/browse/SIMPLY-1047, which could cause people to lose their Adobe IDs as we change from having NYPL's circulation manager handle Adobe IDs, to having the library registry do it.

This branch gives responsibility for the delegate servers to the ShortClientTokenDecoder. When we're asked to decode a short client token, we ask the delegate servers to decode it first, and decode it ourselves if that doesn't work. Either way, we create a DelegatedPatronIdentifier containing the decoded information. This makes sure that if the delegation is removed, the lookup will succeed.

This introduces a number of assumptions that are theoretically problematic (which is why I didn't do things this way in the first place) but in practice should be fine. First, both the library registry and the delegate server need to know about the same libraries with the same short names and shared secrets. Otherwise the library registry will end up with Adobe IDs for libraries it can't actually validate itself.

Second, before this change we could have delegated lookups to any server; now we can only delegate to servers that understand Short Client Tokens, i.e. circulation managers.

These assumptions are fine, because the delegate system is only intended to manage the changeover from the circulation manager to the library registry. If we want to, we can go back to the old system after we switch over.

### DIFF
--- a/model.py
+++ b/model.py
@@ -1562,7 +1562,15 @@ class DelegatedPatronIdentifier(Base):
             patron_identifier=patron_identifier, type=identifier_type
         )
         if is_new:
-            identifier.delegated_identifier = create_function()
+            if callable(create_function):
+                # We are in charge of creating the delegated identifier.
+                delegated_identifier = create_function()
+            else:
+                # We haven't heard of this patron before, but some
+                # other server does know about them, and they told us
+                # this is the delegated identifier.
+                delegated_identifier=create_function
+            identifier.delegated_identifier = delegated_identifier
         return identifier, is_new
 
 
@@ -1583,9 +1591,17 @@ class ShortClientTokenDecoder(ShortClientTokenTool):
         value = "urn:uuid:0" + u[1:]
         return value
 
-    def __init__(self, node_value):
+    def __init__(self, node_value, delegates):
         super(ShortClientTokenDecoder, self).__init__()
+        if isinstance(node_value, basestring):
+            # The node value may be stored in hex form (that's how
+            # Adobe gives it out) or as the equivalent decimal number.
+            if node_value.startswith('0x'):
+                node_value = int(node_value, 16)
+            else:
+                node_value = int(node_value)
         self.node_value = node_value
+        self.delegates = delegates
 
     def decode(self, _db, token):
         """Decode a short client token.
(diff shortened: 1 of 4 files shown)


leonardr:
I think I need to go back to the old behavior for authdata, because that's a whole other weird use case not related to Short Client Tokens.

leonardr:
OK, this should be much better. We ask the delegates every time, whether or not the input is a Short Client Token. If the input _is_ a Short Client Token, we create a DelegatedPatronIdentifier out of the result; otherwise we just pass on the result as is.

model.py
```diff
@@ -1606,20 +1622,60 @@ def decode_two_part(self, _db, username, password):
...
+            # The delegates couldn't help us; let's try to do it
+            # ourselves.
+            try:
+                signature = self.adobe_base64_decode(password)
+            except Exception, e:
+                raise ValueError("Invalid password: %s" % password)
+
+            patron_identifier, account_id = self._decode(
+                _db, username, signature
+            )
```
aslagle: Maybe this could return the library and expires so you don't have to call split_token twice.

model.py
```diff
@@ -1562,7 +1562,15 @@ def get_one_or_create(
...
         if is_new:
-            identifier.delegated_identifier = create_function()
+            if callable(create_function):
+                # We are in charge of creating the delegated identifier.
+                delegated_identifier = create_function()
+            else:
+                # We haven't heard of this patron before, but some
+                # other server does know about them, and they told us
+                # this is the delegated identifier.
+                delegated_identifier=create_function
```
aslagle: This is confusing, could you add another argument instead of using create_function?

model.py
leonardr: I experimented with this idea and it makes the code a lot more complicated for a pretty small performance boost. This is because that first split_token call serves two purposes: if the username isn't actually a Short Client Token, split_token raises an exception which lets us bail out immediately.

It's possible (in fact, likely) that a delegated server could give a response to a username/password that's not a Short Client Token. Without the bailout early in `decode_two_part`, we might send an HTTP request to a foreign server, trigger its side effects, and then discover that we jumped the gun. In that case we'll need to bail out anyway and have the calling code make the HTTP request again.

It's possible there's some design that's altogether different and solves both these problems, but the inefficiency is small enough that I'd prefer to leave it alone.

aslagle:
Is it possible for the account id for a given patron identifier to change?

PR #70, opened 2018-08-01
Generate a wider variety of library short names, and add a duplicate check.

leonardr:
This branch addresses https://jira.nypl.org/browse/SIMPLY-1057, which isn't as bad as I thought when I started the work. But the duplicate check will be really useful.

### DIFF
--- a/controller.py
+++ b/controller.py
@@ -501,10 +501,9 @@ class LibraryRegistryController(object):
             encryptor = PKCS1_OAEP.new(public_key)
 
             if not library.short_name:
-                # TODO: This needs to be base64 encoded or something,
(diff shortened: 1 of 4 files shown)


controller.py
```diff
@@ -501,10 +501,9 @@ def _make_request(url, on_404, on_timeout, on_exception, allow_401=False):
             encryptor = PKCS1_OAEP.new(public_key)
 
             if not library.short_name:
-                # TODO: This needs to be base64 encoded or something,
-                # not hex encoded. Hex encoding limits us to 256*3
-                # libraries. Also we need to handle collisions.
```
leonardr: 256*3 was way off.

PR #72, opened 2018-08-13
Send a problem detail document if a library can't be registered due to a problem with the SMTP server that's supposed to send out the validation email.

leonardr:
This was based on a conversation with Charles Horn of Internet Archive. Our library registry was configured with an email address that Amazon's SMTP server wouldn't accept as the `From:` address. For the time being I've changed the email address we use, but I also made this change so that if this problem happens again, the end-user will get a helpful error message instead of the server just crashing.

### DIFF
--- a/controller.py
+++ b/controller.py
@@ -7,6 +7,7 @@ from flask import (
     url_for,
 )
 import requests
+from smtplib import SMTPException
 import json
 import feedparser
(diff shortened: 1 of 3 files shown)


leonardr:
Yeah, in fact we're still having them -- Charles said he never got the validation emails even though our SMTP server accepted them for delivery.

PR #73, opened 2018-08-14
Allow registration to change an existing library's authentication_url if the registration form submission included the library's shared secret

leonardr:
This partially addresses https://jira.nypl.org/browse/SIMPLY-1031. There will also be a circulation branch that starts sending the shared secret as part of registration.

### DIFF
--- a/controller.py
+++ b/controller.py
@@ -368,8 +368,9 @@ class LibraryRegistryController(object):
             opds_url = auth_document.root['href']
         else:
             failure_detail = _("The OPDS authentication document is missing a 'start' link to the root OPDS feed.")
-        if auth_document.id != auth_response.url:
-            failure_detail = _("The OPDS authentication document's id (%(id)s) doesn't match its url (%(url)s).", id=auth_document.id, url=auth_response.url)
+
+        if auth_document.id != auth_response.final_url:
+            failure_detail = _("The OPDS authentication document's id (%(id)s) doesn't match its url (%(url)s).", id=auth_document.id, url=auth_response.final_url)
         if failure_detail:
             self.log.error(
                 "Registration of %s failed: %s", auth_url, failure_detail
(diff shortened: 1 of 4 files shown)


controller.py
```diff
@@ -367,8 +368,9 @@ def _make_request(url, on_404, on_timeout, on_exception, allow_401=False):
             opds_url = auth_document.root['href']
         else:
             failure_detail = _("The OPDS authentication document is missing a 'start' link to the root OPDS feed.")
-        if auth_document.id != auth_response.url:
-            failure_detail = _("The OPDS authentication document's id (%(id)s) doesn't match its url (%(url)s).", id=auth_document.id, url=auth_response.url)
+
+        if auth_document.id != auth_response.final_url:
+            failure_detail = _("The OPDS authentication document's id (%(id)s) doesn't match its url (%(url)s).", id=auth_document.id, url=auth_response.final_url)
```
leonardr: This handles the case where there's a redirect between the URL provided in the registry submission and the final URL of the authentication document. This required a number of minor changes to the tests, but setting up a redirect is another way to change which URL you're using for a circulation manager.

controller.py
```diff
@@ -525,6 +561,27 @@ def dupe_check(candidate):
...
+        :param library: A Library
+        :param new_authentication_url: A proposed new value for
+            Library.authentication_url
+        :param provided_shared_secret: Allegedly, the library's
+            shared secret.
+        """
+        if library.shared_secret != provided_shared_secret:
+            return AUTHENTICATION_FAILURE.detailed(
+                _("Provided shared secret is invalid")
+            )
```
leonardr: This can't happen the way we call `_update_library_authentication_url` currently; it's just for the sake of paranoia.

tests/test_controller.py
```diff
@@ -444,28 +443,32 @@ def test_register_fails_on_non_authentication_document(self):
...
-        doesn't match the URL it was retrieved from.
-        """
+        # The request returns an authentication document but its `id`
+        # doesn't match the final URL it was retrieved from.
         auth_document = self._auth_document()
-        self.http_client.queue_response(200, content=json.dumps(auth_document))
+        self.http_client.queue_response(
+            200, content=json.dumps(auth_document),
+            url="http://a-different-url/"
+        )
```
leonardr: Of all the tests we had to change to accommodate the new `final_url` checking code, this is the only tricky one. The failure here is that although the final location of the authentication document matches the URL in the registration form submission, it doesn't match the `id` of the authentication document itself.

controller.py
```diff
@@ -426,10 +428,44 @@ def _make_request(url, on_404, on_timeout, on_exception, allow_401=False):
...
+            # Either this is a library at a known authentication URL
+            # or it's a brand new library.
+            library, is_new = get_one_or_create(
+                self._db, Library,
+                authentication_url=auth_url
+            )
+            if opds_url:
+                library.opds_url = opds_url
+
+        if elevated_permissions and library.authentication_url != auth_url:
```
aslagle: Do you need to check the original URL as well as the final URL here for the redirect thing to work? If not, could you explain how it would work?

leonardr: This wouldn't be used in the circulation manager, because the circulation manager always sends the current address to its authentication document. But here's how it could work:

I POST to the registry with url=http://get-an-auth-document/
Registry GETs http://get-an-auth-document/ and is redirected to https://library/my-auth-document/
Registry GETs https://library/my-auth-document/ and gets an auth document.

Without this change, the `id` of the auth document is compared against http://get-an-auth-document/. With this change in place, it's compared against 
https://library/my-auth-document/.

Basically, it's more important to validate the integrity of the auth document itself (which is going into the database) than to check it against a URL that was provided as input into the registration process.

aslagle: I was thinking setting up a redirect would be an alternate way of changing your URL even if you don't have the shared secret. I'm not sure why anyone would need to post an old URL to the registry - if you can change your auth document you can probably change that too.

leonardr: I see what you're saying. There could be a follow-up to this branch that treats a redirect from a known `authentication_url` to a new `authentication_url` as a source of elevated permissions. That would let you change the URL even if you never set up a shared secret. I don't want to do that on this branch, but I'll look into doing this sooner rather than later if you've got a server that needs this feature.

leonardr: Cool, I agree.

controller.py
```diff
@@ -426,10 +428,44 @@ def _make_request(url, on_404, on_timeout, on_exception, allow_401=False):
...
+        if shared_secret:
+            # Look up a library by the provided shared secret. This
+            # will let us handle the case where the library has
+            # changed URLs (auth_url does not match
+            # library.authentication_url) but the shared secret is the
+            # same.
+            library = get_one(
+                self._db, Library,
+                shared_secret=shared_secret
+            )
```
aslagle: Since we are looking up libraries by shared secret now, we might want something like the short name dupe check, even if it's extremely unlikely to be needed.

PR #74, opened 2018-08-16
Add link to auth doc in catalog

adriana-alter:
Resolves https://jira.nypl.org/browse/SIMPLY-1094


aslagle:
You have some test failures but otherwise looks fine.

PR #75, opened 2018-08-20
Added routes for getting a specific library's catalog entry.

aslagle:


leonardr:
In private conversation we agreed that it would be better to drop the separate lookup for libraries in a non-production state. When you're giving out a list of libraries, it makes sense to include or not include non-production libraries. When someone is asking about a specific library they want to know about that library -- its production status is of secondary importance.

controller.py
```diff
@@ -194,6 +194,19 @@ def search(self, ip_address=None, live=True):
             )
             return Response(body, 200, headers)
 
+    def library(self, uuid):
+        library = Library.for_urn(self._db, "urn:uuid:" + uuid)
```
leonardr: The specific place I'd apply the change suggested in my review comment is here -- only add 'urn:uuid' to the incoming `uuid` if it's not already present.

PR #78, opened 2018-10-26
Don't crash when asked to generate an email containing non-ASCII characters

leonardr:
This branch addresses https://jira.nypl.org/browse/SIMPLY-1326

All outgoing emails are MIME multipart emails with a single plain-text part. This is necessary because the MIME envelope is where we explain how we've converted the UTF-8 contents to ASCII. In this case we've done it using the quoted-printable transfer encoding.

We still don't support non-ASCII characters in email addresses or the names of senders or recipients, but that's not as urgent.


leonardr:
Here's the actual email message generated during the 'snowman' test. I wrote it to a file and opened it in Thunderbird to make sure it rendered correctly.

```
Content-Type: multipart/mixed; boundary="===============0348193607107485423=="
MIME-Version: 1.0
From: me@example.com
To: you@example.com
Subject: =?utf-8?q?A_snowman_for_you!_=E2=98=83?=

--===============0348193607107485423==
Content-Type: text/plain; charset="utf-8"
MIME-Version: 1.0
Content-Transfer-Encoding: quoted-printable

Here he is: =E2=98=83
--===============0348193607107485423==--
```

tests/test_emailer.py
```diff
@@ -23,14 +24,47 @@ def test_body(self):
...
+    def test_unicode_quoted_printable(self):
+        # Create an email message that includes Unicode characters in
+        # its subject and body.
+        snowman = u"\N{SNOWMAN}"
+        template = EmailTemplate(
+            u"A snowman for you! %s" % snowman,
+            u"Here he is: %s" % snowman
         )
+        body = template.body("me@example.com", "you@example.com")
+        set_trace()
```
aslagle: You left a set_trace here, but otherwise looks good.

leonardr:
This needs another look. I had to redo the email tests because I can no longer check that the content of an email is a specific string.

PR #79, opened 2018-10-28
If the registry provides access to a web-based client, advertise that fact.

aslagle:
I think the same should be done for mobile clients but I'm not sure how best to identify them.

### DIFF
--- a/config.py
+++ b/config.py
@@ -47,6 +47,11 @@ class Configuration(object):
     # have problems.
     REGISTRY_CONTACT_EMAIL = "registry_contact_email"
 
+    # If the registry provides access to a web-based client, it can
+    # specify the URL with this setting. The URL may be templated and contain
+    # a `{uuid}` expression, to provide the web URL for a specific library.
+    WEB_CLIENT_URL = "web_client_url"
+
     @classmethod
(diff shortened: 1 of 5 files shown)


leonardr:
I'd like to know more about the usage here. In particular, is the client supposed to go to the `web-client` URL and then pick a library, or is the client supposed to pick a library and then go to its `web-client` URL?

If the latter, I would rather see this implemented by giving each individual library in the catalog a link with `rel="http://librarysimplified.org/rel/web-client"`. If template interpolation is necessary, have the library registry do it and save the client the trouble.  If a specific library has its own special "OPDS web view" server, then let that library advertise its server and avoid the risk that a client will render that library's OPDS feeds with the "wrong" client.

leonardr:
We might not even need a new `rel` -- maybe we could say that `rel="self" type="text/html"` means "Web view for this OPDS feed", as opposed to `rel="alternate" type="text/html"`, which means "Web site of the organization running this OPDS server".

aslagle:
That's true, it can be in the library's registry entry instead.

This is primarily meant to be used by the circulation manager, so it can make links to the web client once it's registered.

PR #80, opened 2018-10-30
Speed up and improve reliability for Place.lookup_inside

leonardr:
Currently `Place.lookup_inside` uses comparison of shapefiles to look up a place inside another place (such as California within the United States, or Kern County within California, or 93203 within Kern County). This is very powerful but it also has big problems.

* It's very slow.

* It fails in common cases. It can't parse "New York, New York" because it can't tell whether the first "New York" is referring to New York City or New York State. It can't parse "Santa Barbara, CA" because California has both a city and a county called Santa Barbara. (In real life, anyone wanting to refer to the county would say "Santa Barbara County, CA".)

* Most of the power is for cases that no real person would use. We wouldn't say "93203, Kern County, California", because "93203" is unambiguous on its own. We wouldn't say "Bakersfield, Kern County, California" -- we would use city and state, the way we do when addressing a letter. We wouldn't say "Chicago, US" because it's redundant.

* You can look up a smaller place inside a larger one (e.g. the United States within California), because it compares the intersection of shapefiles, rather than checking that one contains the other.

This branch changes what it means for a place to be "inside" another place for purposes of `Place.lookup_inside`. Now, place A is inside place B only if `A.parent == B`. It works because we set up `.parent` to correspond to the way real people think about places. In particular, the parent of a city is its state, not its county. (The parent of a postal code is its state, and I think we might need to change that to be the nation so that you can just say "93203" instead of "93203, CA".)

The old behavior is preserved in case we need it for anything, but I think we'll find that we don't.

This is work towards https://jira.nypl.org/browse/SIMPLY-1044 but I don't think it entirely resolves that issue.

### DIFF
--- a/model.py
+++ b/model.py
@@ -1152,13 +1152,29 @@ class Place(Base):
         touches = func.ST_Touches(Place.geometry, self.geometry)
         return qu.filter(intersects).filter(touches==False)
 
-    def lookup_inside(self, name):
-        """Look up a named Place that geographically intersects this Place.
+    def lookup_inside(self, name, using_overlap=False):
+        """Look up a named Place that is geographically 'inside' this Place.
+
+        :param name: The name of a place, such as "Boston" or
+        "Calabasas, CA", or "Cook County".
+
+        :param using_overlap: If this is true, then place A is
+        'inside' place B if their shapes overlap, not counting
+        borders. For example, Montgomery is 'inside' Montgomery
+        County, Alabama, and the United States. However, Alabama is
+        not 'inside' Georgia (even though they share a border).
+
+        If `using_overlap` is false, then place A is 'inside' place B
+        only if B is the .parent of A. In this case, Alabama is
+        considered to be 'inside' the United States, but Montgomery is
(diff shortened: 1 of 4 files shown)


leonardr:
I added some code that lets you look up postal codes within their parent or their grandparent, which solves the "93203" problem without making us change the dataset. It might still be worth changing the dataset, because nobody is going to type "93203, CA", but we don't have to do it.

aslagle:
If you used the grandparent technique for cities as well as postal codes you could handle city-county lookups as well as city-state, though I agree it's unlikely anyone would use that.

aslagle:
No one would type "93203, CA", but maybe they would type "Arvin, CA 93203" since that's what the postal service likes.

leonardr:
These are good ideas, we can revisit once we see how people actually use it. The main problem I'm seeing is that in the vast majority of cases coverage area can be represented as a simple string, but we can't parse that string because there's no indication of which country "Los Angeles, CA" is in. I'm going to work on that next.

PR #81, opened 2018-10-30
Allow registries to define a default nation which is used when processing incoming place names that don't mention one

leonardr:
According to https://github.com/NYPL-Simplified/Simplified/wiki/Authentication-For-OPDS-Extensions#service_area, a place name like "San Francisco, CA" is invalid. But for the vast majority of libraries, that's the simplest way to represent the library's service area.

I made up that format and I can change the spec a bit (and probably will) to make things easier for people, but I can't change the fact that "San Francisco, CA" doesn't explicitly say which _country_ we're talking about. I don't want to add US-centric bits to the spec, but the vast majority of the people using NYPL's library registry will be in the US, and it would be nice to make things easier for them.

So, this branch allows a library registry to define a 'default nation'. If a coverage area comes in and it doesn't mention a nation, we try out the assumption that all those places are in the default nation _for that registry_. If that makes sense, everything's fine. Otherwise, we send an error, but we would have sent an error anyway.

### DIFF
--- a/authentication_document.py
+++ b/authentication_document.py
@@ -173,52 +173,52 @@ class AuthenticationDocument(object):
             coverage = dict() # Do no more processing
 
         elif not isinstance(coverage, dict):
-            # The coverage is not in { country: place } format.
-            # Convert it into that format using the default country.
-            default_country = place_class.default_country(_db)
-            if default_country:
-                coverage = {default_country.abbreviated_name : coverage }
+            # The coverage is not in { nation: place } format.
+            # Convert it into that format using the default nation.
+            default_nation = place_class.default_nation(_db)
+            if default_nation:
+                coverage = {default_nation.abbreviated_name : coverage }
             else:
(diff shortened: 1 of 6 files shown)


authentication_document.py
```diff
@@ -168,42 +170,55 @@ def parse_coverage(cls, _db, coverage, place_class=Place):
...
+            if default_nation:
+                coverage = {default_nation.abbreviated_name : coverage }
+            else:
+                # Oops, that's not going to work. We don't know which
+                # nation this place is in. Return a coverage object
+                # that makes it semi-clear what the problem is.
+                unknown["??"] = coverage
+                coverage = dict() # Do no more processing
 
-        for country, places in coverage.items():
```
leonardr: The rest of this code is just consistently using the term 'nation' instead of 'country'.
We say 'country' a couple other places in the code, and I'd eventually like to change it everywhere to match the constant we use for `Place.type`.

PR #83, opened 2018-11-01
Advertise each library's focus and eligibility areas in its catalog entry

leonardr:
This branch adds links to a library's OPDS catalog which let a client grab GeoJSON representations of its focus area and eligibility area. This can be used either by a circulation manager's admin interface ("did I get the service area right?") or by a mobile or web client ("is this my local library?")

I added two controllers which take a library UUID as their input, so rather than copy the code in the existing `library` controller (for looking up a specific library's catalog entry), I wrote the equivalent of the `has_library` function we use in the circulation manager.

Unlike that `has_library`, though, this one is testable, because I created a factory function which takes an app server as an argument and generates a `has_library` implementation for that app server.

### DIFF
--- a/tests/test_app.py
+++ b/tests/test_app.py
@@ -20,7 +20,7 @@ class TestAppHelpers(ControllerTest):
             return "Called with library %s" % flask.request.library.name
 
         def assert_not_found(uuid):
-            response = route_function()
+            response = route_function(uuid)
             eq_(LIBRARY_NOT_FOUND.response, response)
 
         assert_not_found(uuid=None)
--- a/tests/test_app.py
+++ b/tests/test_app.py
@@ -1,0 +1,37 @@
+import flask
+from app_helpers import has_library_factory
+from nose.tools import (
+    eq_,
+    set_trace,
+)
+from problem_details import (
+    LIBRARY_NOT_FOUND
+)
+from test_controller import ControllerTest
+from testing import DatabaseTest
+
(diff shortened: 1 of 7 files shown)


tests/test_controller.py
```diff
@@ -403,23 +436,11 @@ def test_search_qa(self):
...
-            [catalog_entry] = json.loads(response.data).get("catalogs")
-            eq_(nypl.name, catalog_entry.get("metadata").get("title"))
-            eq_(nypl.internal_urn, catalog_entry.get("metadata").get("id"))
-
-            # Or its UUID without the prefix.
-            uuid = nypl.internal_urn[len("urn:uuid:"):]
-            response = self.controller.library(uuid)
-            [catalog_entry] = json.loads(response.data).get("catalogs")
-            eq_(nypl.name, catalog_entry.get("metadata").get("title"))
-            eq_(nypl.internal_urn, catalog_entry.get("metadata").get("id"))
```
leonardr: These tests were moved into TestBaseController

tests/test_app.py
```diff
@@ -0,0 +1,37 @@
...
+
+    def test_has_library(self):
+        has_library = has_library_factory(self.app)
+
+        @has_library
+        def route_function():
+            return "Called with library %s" % flask.request.library.name
+
+        def assert_not_found(uuid):
+            response = route_function()
```
aslagle: Should the uuid be passed in here?

PR #86, opened 2018-11-02
It's now possible to simulate a request from anywhere by passing in latitude/longitude coordinates.

leonardr:
This branch defines a new annotator called `uses_location`. A routing function annotated with `uses_location` should take an argument called `_location`. The app server will make its best guess at the user's location and pass in that Point as `_location`.

This has two benefits:

* You can put `_location=latitude,longitude` in any request to simulate a request from that point. If incoming `_location` is specified in the request, `uses_location` will use that point instead of trying to figure it out from the IP address. This lets us simulate a request from anywhere when testing.
* Location-aware controllers now take a Point as input rather than an IP address, making them easier to test.

### DIFF
--- a/tests/test_util.py
+++ b/tests/test_util.py
@@ -14,3 +14,17 @@ class TestGeometryUtility(object):
         point = GeometryUtility.point_from_ip("65.88.88.124")
         eq_('SRID=4326;POINT (-73.9813 40.7769)', point)
 
+    def test_point_from_string(self):
+        m = GeometryUtility.point_from_string
+
+        # Lots of strings don't map to latitude/longitude.
+        eq_(None, m(None))
+        eq_(None, m("No comma"))
+        eq_(None, m("Not a number, -71"))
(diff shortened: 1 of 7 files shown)


leonardr:
This resolves https://jira.nypl.org/browse/SIMPLY-1380.

tests/test_util.py
```diff
@@ -14,3 +14,17 @@ def test_point_from_ip(self):
...
+
+        # Lots of strings don't map to latitude/longitude.
+        eq_(None, m(None))
+        eq_(None, m("No comma"))
+        eq_(None, m("Not a number, -71"))
+        eq_(None, m("-400,1"))
+        eq_(None, m("1,400"))
+
+        # Here are some strings that do.
+        for coords in ("40.7769, -73.9813", "40.7769,-73.9813"):
```
aslagle: Are you testing the same coordinates twice?

aslagle: Oh, one has a space.

PR #90, opened 2019-01-04
Configure front end

adriana-alter:


controller.py
```diff
@@ -215,6 +256,51 @@ def search(self, location, live=True):
...
+                    short_name=library.short_name,
+                    uuid=uuid,
+                    registry_stage=library.registry_stage,
+                    library_stage=library.library_stage,
+                )]
+        return dict(libraries=libraries)
+
+    def library_details(self, uuid):
+        # Return complete information about one specific library; this is triggered when an admin opens the detail page for a library.
+        library = self.library_for_request(uuid)
```
leonardr: Wherever you call `library_for_request` you need to check to see if you got a ProblemDetail, and return the ProblemDetail immediately if so.

aslagle:
What's the plan for authentication?

adriana-alter:
@aslagle The plan for authentication is to just create one account (with hard-coded credentials), since the only people who will be using this in the immediately-foreseeable future are you and Leonard; see https://jira.nypl.org/browse/SIMPLY-1574

package.json
```diff
@@ -0,0 +1,25 @@
...
+  "name": "library_registry",
+  "version": "1.0.0",
+  "description": "After creating the database for this application, you must enable two PostgreSQL extentions:",
+  "main": "index.js",
+  "directories": {
+    "test": "tests"
+  },
+  "scripts": {
+    "test": "echo \"Error: no test specified\" && exit 1"
+  },
```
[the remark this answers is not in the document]
adriana-alter: Should license be "Apache-2.0" (like it is on `circulation`)?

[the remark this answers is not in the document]
aslagle: Yes

controller.py
```diff
@@ -215,6 +256,55 @@ def search(self, location, live=True):
...
+            authentication_url=library.authentication_url,
+            online_registration=library.online_registration,
+            description=library.description,
+            timestamp=library.timestamp,
+            internal_urn=library.internal_urn,
+            library_stage=library._library_stage,
+            opds_url=library.opds_url,
+            registry_stage=library.registry_stage,
+            web_url=library.web_url,
+        )
```
aslagle: It would be good to have contact emails in this info as well.

controller.py
```diff
@@ -215,6 +256,55 @@ def search(self, location, live=True):
...
+    def edit_registration(self):
+        # Edit a specific library's registry_stage and library_stage based on information which an admin has submitted in the interface.
+        uuid = flask.request.form.get("uuid")
+        library = self.library_for_request(uuid)
+        if isinstance(library, ProblemDetail):
+            return library
+        registry_stage = flask.request.form.get("Registry Stage")
+        library_stage = flask.request.form.get("Library Stage")
+
+        library._library_stage = library_stage
```
aslagle: This is giving me a 500 error if I change library stage for a production library - "This library is already in production, only the registry can take it out of production."

Should the library stage even be editable by the registry admin?

adriana-alter: Yes, library stage should be editable by the registry admin.  I'm not getting the error--do you have specific steps to reproduce (beyond just setting the library stage to production and then trying to change it to something else)?

controller.py
adriana-alter: Where would we be getting that information?  I'm not seeing anything in the Library model that looks like a contact email...

controller.py
aslagle: You have to set both library and registry stage to production, then try to change the library stage but not the registry stage.

Why should library stage be editable?

controller.py
aslagle: I think they're Hyperlinks that are associated with the library.

adriana-alter: Yes, that seems right--thanks!

controller.py
aslagle: Library stage is the stage the library itself wants to be in, and should come from what the library set on the circ manager. Maybe the registry admin would want to fix it if the library was having trouble, but in general the library should change it on the circ manager and re-register. The registry stage always needs to be updated by registry admins. 

I tried this on Friday right before I wrote the comment.

adriana-alter: I think the point would be to let the admin fix it if the library was having trouble.  (The user base is comprised of you and Leonard; it's not like we have to worry about rogue admins going in and messing up the library stage for no good reason!)

Would you mind trying it again?  I definitely played around with the relevant code on Friday, but I don't remember exactly what time...

leonardr: The main case I was thinking of here is where the circulation manager just shuts down. (This has happened.) At that point we need to change both states to 'cancelled' and the normal rules about what's allowed don't apply.

But I agree that in most cases it doesn't make sense for the registry to modify the library_stage.

aslagle: Tried again and it is working now.

app.py
```diff
@@ -21,6 +21,7 @@
 
 app = Flask(__name__)
 babel = Babel(app)
+app.secret_key = '\x10x\x1a\xc4\x0ef6\xbdcy\x0b\xba\xf8KV\x19\x02\x82\xed\x9b;\xca9#'
```
leonardr: This code is public; we can't put a secret key in here. You can pull it from a sitewide configuration setting instead.

This whole login thing looks like something that should be in a separate branch. I would be okay with merging revision cbc38cea00147b9c7c5cea6d59c74626d966b6e4 and addressing the login bit separately.

PR #91, opened 2019-01-22
Categorize libraryinfo

adriana-alter:
Sorts information about a library into categories before sending it to the front end (so that the interface can more easily organize the information in a user-friendly way).

### DIFF
--- a/tests/test_controller.py
+++ b/tests/test_controller.py
@@ -247,12 +247,12 @@ class TestLibraryRegistryController(ControllerTest):
                 expected_time = [expected_ts.year, expected_ts.month, expected_ts.day]
                 eq_(actual_time, expected_time)
             elif k == "contact_email":
-                expected_contact_email = expected.__dict__.get("name") + "@library.org"
+                expected_contact_email = expected.name + "@library.org"
                 eq_(flattened.get("contact_email"), expected_contact_email)
             elif k == "online_registration":
-                eq_(flattened.get("online_registration"), str(expected.__dict__.get("online_registration")))
+                eq_(flattened.get("online_registration"), str(expected.online_registration))
             else:
-                eq_(flattened.get(k), expected.__dict__.get(k))
+                eq_(flattened.get(k), getattr(expected, k))
 
     def _check_keys(self, library):
         # Helper method to check that the controller is sending the right pieces of information about a library.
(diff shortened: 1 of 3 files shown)


tests/test_controller.py
```diff
@@ -230,80 +230,75 @@ def setup(self):
 
     def _is_library(self, expected, actual):
         # Helper method to check that a library found by a controller is equivalent to a particular library in the database
-        for k in actual.keys():
+
+        flattened = {}
+        for category in ["basic_info", "urls_and_contact", "stages"]:
+            for k in actual[category].keys():
+                flattened[k] = actual[category][k]
```
leonardr: I think you can replace this with this simpler code:

```
flattened = {}
for subdictionary in actual.values():
    flattened.update(subdictionary)
```

tests/test_controller.py
```diff
@@ -230,80 +230,75 @@ def setup(self):
...
-        actual_names = [library.get("name") for library in libraries]
+        expected_names = [library.name for library in [ct, ks, nypl]]
+        actual_names = [library.get("basic_info").get("name") for library in libraries]
         eq_(set(expected_names), set(actual_names))
 
-        self._is_library(self.connecticut_state_library, libraries[0])
-        self._is_library(self.kansas_state_library, libraries[1])
-        self._is_library(self.nypl, libraries[2])
+        for idx, expected in enumerate([ct, ks, nypl]):
+            self._is_library(expected, libraries[idx])
```
leonardr: I think this would be easier to understand:

```
self._is_library(expected[0], ct)
self._is_library(expected[1], ks)
self._is_library(expected[2], nypl)
```

tests/test_controller.py
```diff
@@ -299,7 +301,7 @@ def test_library_details(self):
         eq_(uuid, response.get("uuid"))
 
         self._check_keys(response)
-        self._is_library(library, response)
+        self._is_library(self.nypl, response)
```
leonardr: I think the old code is better -- this will cause `nypl` to be created again, when it doesn't look like that is necessary.

adriana-alter: The old code suddenly started failing--don't know why...  It's still instantiating `library` correctly, but then by the time it's ready to call `_is_library`, `library.__dict__` suddenly contains a sqlalchemy InstanceState object and nothing else.

PR #94, opened 2019-01-30
Redirect to log in form

adriana-alter:


EdwinGuzman:
No :(
I'm not sure what the right way to do this in Python/Flask would be (this is also a new incognito window):
screen shot 2019-01-30 at 5 08 02 pm

adriana-alter:
@EdwinGuzman Try this one: https://github.com/NYPL-Simplified/library_registry/pull/95

EdwinGuzman:
Yes, that fixes it, but why not just update this PR instead of creating a new one? I don't know if that's a good way to fix it in flask so it'd be good to get another opinion.

EdwinGuzman:
Alright. I'll close this one then since you have two.

PR #98, opened 2019-02-11
Email

adriana-alter:
This branch addresses https://jira.nypl.org/browse/SIMPLY-1573.

### DIFF
--- a/controller.py
+++ b/controller.py
@@ -249,11 +249,13 @@ class LibraryRegistryController(BaseController):
         if isinstance(library, ProblemDetail):
             return library
 
-        # Find the resource having to do with the contact email for this library;
-        # we'll need this to look up both the email address and the validation information.
-        resource = self._get_resource(library.id, Hyperlink.INTEGRATION_CONTACT_REL)
-        contact_email = self._contact_email(resource)
-        validated_at = self._validated_at(resource)
+        contact_email = None
+        validated_at = None
+
+        hyperlink = Library.get_hyperlink(library, Hyperlink.INTEGRATION_CONTACT_REL)
+        if hyperlink and hyperlink.resource:
+            contact_email = self._contact_email(hyperlink.resource)
+            validated_at = self._validated_at(hyperlink.resource)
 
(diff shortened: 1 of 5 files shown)


controller.py
```diff
@@ -273,18 +276,52 @@ def library_details(self, uuid, library=None):
...
+            validated_at = validation.started_at
+        return validated_at
+
+    def _get_resource(self, library_id, rel):
+        hyperlink = self._get_hyperlink(library_id, rel)
+        if hyperlink:
+            query = self._db.query(Resource).filter(
+                Resource.id==hyperlink.resource_id
             )
+            return query.first()
```
leonardr: This whole thing can become `hyperlink.resource`.

controller.py
```diff
@@ -273,18 +276,52 @@ def library_details(self, uuid, library=None):
...
 
-        if query.first():
-            contact_email = query.first().href.split("mailto:")[1]
+    def _get_hyperlink(self, library_id, rel):
+        query = self._db.query(Hyperlink).filter(
+            Hyperlink.library_id==library_id,
+            Hyperlink.rel==rel
+        )
+        hyperlink = query.first()
+        return hyperlink
```
leonardr: This code is fine but it might be better to make it a method of the Library class, `Library.hyperlink(rel)`. That way other bits of code can use it.

tests/test_controller.py
```diff
@@ -383,6 +385,48 @@ def test_edit_registration_with_override(self):
...
+    def test_email(self):
+        nypl = self.nypl
+        uuid = nypl.internal_urn.split("uuid:")[1]
+        with self.app.test_request_context("/", method="POST"):
+            flask.request.form = MultiDict([
+                ("uuid", uuid),
+            ])
+            response = self.controller.email()
+        eq_(response.response, [nypl.internal_urn])
+        eq_(response.status_code, 200)
```
leonardr: This needs to also check that an email was actually sent. You should be able to access self.controller.emailer.sent_out and see the email in there.

PR #100, opened 2019-02-15
Manually validate email address instead of sending confirmation link

adriana-alter:


controller.py
```diff
@@ -297,15 +298,14 @@ def email(self):
...
-            )
-        return Response(unicode(library.internal_urn), 200)
+        validation, is_new = get_one_or_create(self._db, Validation, resource=hyperlink.resource)
+        validation.restart()
+        validation.mark_as_successful()
+
+        if (validation.success):
+            return Response(unicode(library.internal_urn), 200)
+        else:
+            return CANNOT_VALIDATE
```
leonardr: Looking at the code I don't see how validation.success would ever be false here, so the code could be removed without hurting anything. Plus, leaving it in would mean testing it, which would be annoying.

PR #102, opened 2019-02-19
Return library details on validate email

adriana-alter:
This change is just to improve the performance on the front end; it makes the library info update itself in a more visually smooth way after an email address is validated.  Goes with https://github.com/NYPL-Simplified/registry_admin/pull/9, which resolves https://jira.nypl.org/browse/SIMPLY-1725.


EdwinGuzman:
This looks fine to me but check the front-end PR for more comments.

PR #104, opened 2019-03-13
Move the library registration code into a standalone class

leonardr:
This branch is almost entirely refactoring. The motivating case is https://jira.nypl.org/browse/SIMPLY-1575. I want to be able to run most of the registration process (download the authentication document, validate it, make changes to the local database) from a script, for every library in the system, without waiting for the administrator of that library to re-click the 'register' button.

This will let us automatically keep up to date on things like library descriptions, logos, etc., without forcing people to remember to re-register whenever they change something. Big changes like moving the server to another URL, or changing the contact email, _will_ require the library administrator to re-click the 'register' button.

The refactoring is basically that I took a big chunk of code out of the LibraryRegistryController and put it in a new class, LibraryRegistrar. I didn't change the tests except for the ones I could move wholesale out of the controller test module and into the registrar test module.

The refactoring triggered a bug in our use of SQLAlchemy that caused tests to fail on Travis but not on my laptop; I'll explain the bug inline.


controller.py
```diff
@@ -460,231 +408,34 @@ def register(self, do_get=HTTP.debuggable_get):
...
-        hyperlinks_to_create = []
+
+        registrar = LibraryRegistrar(self._db, do_get=do_get)
+        result = registrar.register(
+            auth_url, shared_secret, library_stage
+        )
+        if isinstance(result, ProblemDetail):
+            return result
+        (library, library_is_new, from_shared_secret, auth_document,
+         hyperlinks_to_create) = result
```
leonardr: The main downside of this branch is the ugliness of the return value from `LibraryRegistrar.register`. There's a bunch of stuff that's determined during the course of registration, but which is only relevant to the controller. I'd like to hear your thoughts on how to improve this. The only thing I can think of is making a special `RegistrationStatus` class to contain this information, which isn't much better.

registrar.py
```diff
@@ -0,0 +1,440 @@
...
+
+class LibraryRegistrar(object):
+    """Encapsulates the logic of the library registration process."""
+
+    def __init__(self, _db, do_get=HTTP.debuggable_get):
+        self._db = _db
+        self.do_get = do_get
+        self.log = logging.getLogger("Library registrar")
+
+    def reregister(self, library):
```
leonardr: `reregister`, and its test, are the only new code in this branch.

model.py
```diff
@@ -1516,9 +1516,14 @@ def notify(self, emailer, url_for):
...
-            _db, Validation, resource=resource
-        )
+        # Make sure there's a Validation object associated with this
+        # Resource.
+        if resource.validation is None:
+            resource.validation, is_new = create(_db, Validation)
+        else:
+            is_new = False
+        validation = resource.validation
+
```
leonardr: This is the bug fix. The `validations` table doesn't have a `resource_id`. Instead the one-to-one relationship goes the other way: `resources.validation_id`. This means that this `get_one_or_create` call isn't reliable -- "find the Validation for this Resource" is complicated in a way that "find the Resource for this Validation" wouldn't be.

I never figured out why this particular refactoring caused the bug to show up, but I'm happy with this change.

testing.py
```diff
@@ -95,8 +95,6 @@ def setup(self):
         self.longitude_counter = -90
 
     def teardown(self):
-        self._db.rollback()
-
```
leonardr: This rollback was only needed in a single test -- one that causes a database integrity error -- and it seemed weird to roll back the entire database session and then roll back the transaction. It didn't solve the problem I was having but I decided to keep the fix.

tests/test_controller.py
```diff
@@ -1120,9 +1120,12 @@ def send(self, *args, **kwargs):
...
             response = self.controller.register(do_get=self.http_client.do_get)
 
-        # We get back a ProblemDetail.
+        # We get back a ProblemDetail the first time
+        # we got a problem sending an email. In this case, it was
+        # trying to contact the library's 'help' address included in the
+        # library's authentication document.
         eq_(INTEGRATION_ERROR.uri, response.uri)
-        eq_("SMTP error while sending email to mailto:me@library.org",
+        eq_("SMTP error while sending email to mailto:help@library.org",
```
leonardr: This changed because the controller previously pulled the contact address out of the form submission before running the registration code, so that address (me@library.org) was the first one to fail validation. Now the controller pulls the contact address after running the registration code (this is just to make the code nicer), so the help address (gathered by the registration code) is the first one to fail validation.

tests/test_opds.py
```diff
@@ -165,9 +165,8 @@ def test__hyperlink_args(self):
         eq_(dict(href=hyperlink.href, rel=hyperlink.rel), m(hyperlink))
 
         # Create a Validation.
-        validation, is_new = get_one_or_create(
-            self._db, Validation, resource=hyperlink.resource
-        )
+        validation, is_new = create(self._db, Validation)
+        hyperlink.resource.validation = validation
```
leonardr: Here's another case where we were using get_one_or_create in a situation that was (sometimes) a little too tricky for SQLAlchemy to figure out what we were looking for.

controller.py
aslagle: I think the controller could be responsible for looking up or creating the library, so the registrar doesn't have to deal with the shared secret. I think it would be okay for that to happen before attempting to get the auth document, since it can get rolled back later if necessary.

authentication_document.py
```diff
@@ -421,7 +421,6 @@ def _update_service_areas(cls, library, areas, type, service_areas):
                 msgs.append(str(_("The following service area was unknown: %(service_area)s.", service_area=json.dumps(unknown))))
             if ambiguous:
                 msgs.append(str(_("The following service area was ambiguous: %(service_area)s.", service_area=json.dumps(ambiguous))))
-            _db.rollback()
```
leonardr: This session-level rollback was interfering with, and redundant with, the subtransaction rollback that now happens when registration of a new library fails for _any_ reason.

controller.py
```diff
@@ -409,14 +409,70 @@ def register(self, do_get=HTTP.debuggable_get):
...
+            # Look up a library by the provided shared secret. This
+            # will let us handle the case where the library has
+            # changed URLs (auth_url does not match
+            # library.authentication_url) but the shared secret is the
+            # same.
+            library = get_one(self._db, Library, shared_secret=shared_secret)
+            if not library:
+                return AUTHENTICATION_FAILURE.detailed(
+                    _("Provided shared secret is invalid")
+                )
```
leonardr: We now check the shared secret immediately and return a problem detail if it's wrong; previously we were ignoring a bad shared secret unless the request did something that required elevated permissions.

registrar.py
```diff
@@ -414,28 +352,3 @@ def _required_email_address(cls, uri, problem_title):
             problem.title = problem_title
             return problem
         return uri
-
-    @classmethod
-    def _update_library_authentication_url(
-            cls, library, new_authentication_url,
-            new_opds_url, provided_shared_secret,
-    ):
```
leonardr: This method doesn't work anymore because the authentication_url is modified before the call to register(), and the opds_url is modified afterwards.

PR #105, opened 2019-03-14
Endpoint to return all libraries not in a cancelled stage in OPDS format

EdwinGuzman:
Fixes [SIMPLY-1775](https://jira.nypl.org/browse/SIMPLY-1775).

### DIFF
--- a/app.py
+++ b/app.py
@@ -105,6 +105,11 @@ def confirm_resource(resource_id, secret):
         resource_id, secret
     )
 
+@app.route('/libraries')
+@returns_problem_detail
+def libraries_opds():
+    return app.library_registry.registry_controller.libraries_opds()
+
 @app.route('/admin/log_in', methods=["POST"])
 @returns_problem_detail
 def log_in():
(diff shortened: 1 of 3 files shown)


tests/test_controller.py
```diff
@@ -297,6 +297,53 @@ def test_libraries(self):
...
+            name="Test Cancelled Library",
+            short_name="test_cancelled_lib",
+            library_stage=Library.CANCELLED_STAGE,
+            registry_stage=Library.TESTING_STAGE
+        )
+        response = self.controller.libraries()
+        libraries = response.get("libraries")
+
+        # There are currently four libraries,
+        # These libraries have been added from the previous test.
```
leonardr: They're added to every test by data_setup(), which is called by setup(), not left over from the previous test. Tests run in a subtransaction which is rolled back at the end of the test, so they're always separated.

EdwinGuzman: Ah got it.

app.py
```diff
@@ -120,6 +120,11 @@ def log_out():
 def libraries():
     return app.library_registry.registry_controller.libraries()
 
+@app.route('/admin/libraries_opds')
+@returns_problem_detail
+def libraries_opds():
+    return app.library_registry.registry_controller.libraries_opds()
+
```
leonardr: This is talking to SimplyE, not the admin interface, so we don't want `/admin/` in the route. Just routing it from `'/libraries` should be fine.

PR #106, opened 2019-03-14
Script to reregister all libraries

leonardr:
This branch implements the script necessary for https://jira.nypl.org/browse/SIMPLY-1575. It basically sweeps over all non-cancelled libraries and updates our records based on what their authentication documents say.

Since almost all the work is done inside LibraryRegistrar.reregister(), this branch focuses on improving `LibraryScript` so that you can write a script that will either operate on a single named library or on all libraries.

### DIFF
--- a/tests/test_scripts.py
+++ b/tests/test_scripts.py
@@ -16,6 +16,8 @@ from model import (
     create,
     get_one,
 )
+from problem_details import INVALID_INTEGRATION_DOCUMENT
+from registrar import LibraryRegistrar
 from scripts import (
     AddLibraryScript,
     ConfigureEmailerScript,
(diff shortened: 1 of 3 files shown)


tests/test_scripts.py
```diff
@@ -286,6 +331,72 @@ def test_add_settings(self):
         eq_(expect_output, output.getvalue())
 
 
+class TestRegistrationRefreshScript(DatabaseTest):
+
+    def test_run(self):
+        # Verify that run() instantiates a LibraryRegistrar using .registrar,
+        # then calles its reregister() method on every library that it's
```
EdwinGuzman: A very minor typo in `calles`.

EdwinGuzman (approved):

PR #107, opened 2019-04-02
Uszipcode on demand

leonardr:
This branch resolves https://jira.nypl.org/browse/SIMPLY-1760.

The improvement turns out to be pretty marginal when the database is fully stocked with census data and ZIP code data, but this makes it a little better, and it will create a _dramatic_ improvement in the (unlikely) case where a library registry is stocked only with ZIP code data.


model.py
```diff
@@ -1306,6 +1320,44 @@ def lookup_inside(self, name, using_overlap=False):
...
+        will be a Place representing one of the city's postal codes.
+
+        :return: A Place, or None if the lookup fails.
+        """
+        if self.type != Place.STATE:
+            # uszipcodes keeps track of places in terms of their state.
+            return None
+
+        _db = Session.object_session(self)
+        search = uszipcode.SearchEngine(simple_zipcode=True)
```
leonardr: `simple_zipcode=True` means the data download (which happens once every couple of weeks) is a few megabytes instead of half a gigabyte.

EdwinGuzman: How do you track or implement that it downloads just once every couple of weeks?

leonardr: It's internal to `uszipcode`, I don't know for sure that that's the timeframe -- just going on what it says in the docs.

PR #116, opened 2019-05-24
Make libraries endpoint return only production libraries, and add QA

adriana-alter:


controller.py
```diff
@@ -229,6 +229,15 @@ def search(self, location, live=True):
...
     def libraries(self):
+        # Return a specific set of information about all libraries in production; this generates the library list in the admin interface,
+        libraries = []
+        in_production = self._db.query(Library).filter(Library.registry_stage=="production").order_by(Library.name)
+        for library in in_production:
+            uuid = library.internal_urn.split("uuid:")[1]
+            libraries += [self.library_details(uuid, library)]
+        return dict(libraries=libraries)
+
+    def libraries_qa(self):
```
leonardr: Elsewhere in this web app we define a single controller method that takes a `live` argument which toggles whether to look at all libraries or only production libraries. We add two routes to the same method with different arguments. Take a look at `nearby` and `nearby_qa` in `app.py` for an example.

This guarantees that we always build the library list using the same code -- only the query that chooses libraries is different.

controller.py
```diff
@@ -228,23 +228,20 @@ def search(self, location, live=True):
...
-        all = self._db.query(Library).order_by(Library.name)
-        for library in all:
+    def libraries(self, live=True):
+        # Return a specific set of information about all libraries in production; this generates the library list in the admin interface.
+        # If :param live is set to False, libraries in testing will also be shown.
+        result = []
+        libraries = self._db.query(Library).order_by(Library.name)
+
+        if live:
+            libraries = libraries.filter(Library.registry_stage=="production")
```
leonardr: We have a constant for this -- `Library.PRODUCTION_STAGE`.

PR #128, opened 2019-07-17
Enable admins to input/edit and view each library's PLS ID via the in…

adriana-alter:
…terface


tests/test_controller.py
```diff
@@ -550,6 +554,35 @@ def test_missing_email_error(self):
         eq_(response.detail, 'The contact URI for this library is missing or invalid')
         eq_(response.uri, 'http://librarysimplified.org/terms/problem/invalid-contact-uri')
 
+    def test_add_or_edit_pls_id(self):
+        # Test that the user can input a new PLS ID
+        library = self.nypl
+        ConfigurationSetting.for_library("pls_id", library)
```
leonardr: This line doesn't do anything.

controller.py
```diff
@@ -272,14 +272,15 @@ def library_details(self, uuid, library=None):
         hyperlink = Library.get_hyperlink(library, Hyperlink.INTEGRATION_CONTACT_REL)
         contact_email = self._contact_email(hyperlink)
         validated_at = self._validated_at(hyperlink)
-
+        pls_id = ConfigurationSetting.for_library("pls_id", library)
```
leonardr: "pls_id" is a cryptic string that's used a lot; it would be good to keep it as a constant in `Library` with a comment explaining what it's for.

controller.py
```diff
@@ -347,6 +348,15 @@ def edit_registration(self):
         library.registry_stage = registry_stage
         return Response(unicode(library.internal_urn), 200)
 
+    def add_or_edit_pls_id(self):
+        uuid = flask.request.form.get("uuid")
+        library = self.library_for_request(uuid)
+        if isinstance(library, ProblemDetail):
+            return library
```
leonardr: This bit needs a test -- if you pass in an invalid UUID to the controller you get a ProblemDetail.

PR #131, opened 2019-08-06
Improve search

adriana-alter:
Resolves https://jira.nypl.org/browse/SIMPLY-2197.

Improves searching for a library in three ways:

1) Enables searching by part of a library's name (e.g. just typing "Enlow" rather than "Ruth Enlow Library of Garrett County").

2) Enables searching by place name.

3) Enables searching by description.  (This is important because a LOT of libraries have set their geographical info as part of their description rather than via the service/focus areas attributes.)


model.py
```diff
@@ -852,6 +844,37 @@ def search_by_location_name(cls, _db, query, type=None, here=None,
     us_zip_plus_4 = re.compile("^[0-9]{5}-[0-9]{4}$")
     running_whitespace = re.compile("\s+")
 
+    @classmethod
+    def create_query(cls, _db, here=None, production=True, *args):
```
leonardr: Passing in a list of clauses and ORing them together is a nice solution.

PR #134, opened 2019-08-14
Adapt to Python 2 changes made to Python 3 code after running 2to3.

leonardr:
This branch makes our library registry code forward-compatible with Python 3. It's very similar to the other branches of this sort. 

The biggest change is, we found another dependency that doesn't work in Python 3. You can install python-geoip-geolite2 on Python 3 (which is why I didn't discover the problem earlier) but the code won't run. Fortunately maxminddb-geolite2 has picked up where python-geoip-geolite2 left off, and the interface is almost the same.


controller.py
```diff
@@ -322,6 +325,8 @@ def validate_email(self):
         # Manually validate an email address, without the admin having to click on a confirmation link
         uuid = flask.request.form.get("uuid")
         library = self.library_for_request(uuid)
+        if isinstance(library, ProblemDetail):
+            return library
```
leonardr: This won't come up a lot and doesn't relate to the Python 3 conversion, but a problem _inside_ library_for_request was causing it to return a ProblemDetail, and then this method wasn't handling the ProblemDetail properly. I fixed it and added a test.

emailer.py
```diff
@@ -2,15 +2,15 @@
 from email.header import Header
 from email.mime.multipart import MIMEMultipart
 from email.mime.text import MIMEText
-from email import Charset
+from email import charset
```
leonardr: email.charset is a 'legacy' API but I don't understand how the new API works so I'm leaving it alone for now.

log.py
```diff
@@ -34,16 +34,16 @@ def format(self, record):
...
 
-class UTF8Formatter(logging.Formatter):
-    """Encode all Unicode output to UTF-8 to prevent encoding errors."""
+
+class StringFormatter(logging.Formatter):
+    """Encode all output as a string.
+    
+    In Python 2, this means a UTF-8 bytestring. In Python 3, it means a
+    Unicode string.
+    """
```
leonardr: This is a copy of a file from core, so the same changes need to be made here as in core.

tests/test_controller.py
```diff
@@ -519,10 +519,22 @@ def test_edit_registration_with_override(self):
...
+        with self.app.test_request_context("/", method="POST"):
+            flask.request.form = MultiDict([
+                ("uuid", "no:such:library"),
+            ])
+            response = self.controller.validate_email()
+        assert isinstance(response, ProblemDetail)
+        eq_(response.status_code, 404)
+        eq_(response.title, LIBRARY_NOT_FOUND.title)
+        eq_(response.uri, LIBRARY_NOT_FOUND.uri)
+
```
leonardr: Here's the test of the new code I added above.

tests/test_controller.py
```diff
@@ -709,7 +721,7 @@ def test_nearby(self):
             eq_("NYPL", nypl['metadata']['title'])
             eq_("0 km.", nypl['metadata']['distance'])
             eq_("Connecticut State Library", ct['metadata']['title'])
-            eq_("35 km.", ct['metadata']['distance'])
+            eq_("29 km.", ct['metadata']['distance'])
```
leonardr: When we changed GeoIP libraries, we also changed which version of the database we're using, which changed the estimated coordinates assigned to IP addresses. This affects distance and any representations of latitude and longitude.

tests/test_language.py
```diff
@@ -64,4 +64,4 @@ def test_name_for_languageset(self):
         eq_(u"English/español", m(["eng", "spa"]))
         eq_(u"español/English", m("spa,eng"))
         eq_(u"español/English/Chinese", m(["spa","eng","chi"]))
-        assert_raises(ValueError(m, ["eng, nxx"]))
+        assert_raises(ValueError, m, ["eng, nxx"])
```
leonardr: This test wasn't doing anything  because of a misplaced closing parenthesis-- it was basically `assert_raises(ValueError())`. That's not allowed in the Python 3 version of nose -- you have to actually run some code or the assertion will fail.

tests/test_util.py
```diff
@@ -12,7 +12,10 @@ def test_point(self):
 
     def test_point_from_ip(self):
         point = GeometryUtility.point_from_ip("65.88.88.124")
-        eq_('SRID=4326;POINT (-73.9813 40.7769)', point)
+        eq_('SRID=4326;POINT (-73.9169 40.8056)', point)
+
+        point = GeometryUtility.point_from_ip("127.0.0.1")
+        eq_(None, point)
```
leonardr: We had code for what to do if point_for_ip returned `None` but no test for it.

authentication_document.py
```diff
@@ -462,7 +462,7 @@ def _update_collection_size(self, library, sizes):
                     CollectionSummary.set(library, None, unknown_size)
                 )
         except ValueError, e:
-            return INVALID_INTEGRATION_DOCUMENT.detailed(e.message)
+            return INVALID_INTEGRATION_DOCUMENT.detailed(unicode(e))
```
EdwinGuzman: This should be the same as `str(e)` which we've been doing in `circulation`, right?

emailer.py
EdwinGuzman: Is it being `legacy` related to the case change?

authentication_document.py
leonardr: I have given this some thought and I think either one is going to work. In problem_detail.py, when the problem detail is converted to JSON, almost the first thing we do is convert the detail to Unicode. When we convert to Python 3, str(e.message) and unicode(e.message) will both end up as str(e.message), meaning a Unicode string.

tests/test_controller.py
EdwinGuzman: That's unfortunate. Is this something to worry about for other data?

leonardr: I don't think so, the differences are approximately the distance between Manhattan and the Bronx. Not something that would affect the way we use the estimated locations.

emailer.py
leonardr: Sort of -- in Python 2, `Charset` is a class, and in Python 3 it's been moved to a module called `charset`. But they've moved other classes to modules without also deprecating them.

util/string_helpers.py
```diff
@@ -0,0 +1,86 @@
...
+def random_string(size):
+    """Generate a random string of binary, encoded as hex digits.
+
+    :param: Size of binary string in bytes.
+    :return: A Unicode string.
+    """
+    return binascii.hexlify(os.urandom(size)).decode("utf8")
+
+
+def native_string(x):
```
EdwinGuzman: Is this something that could also be used in circulation/core? It'd be nice to share this file between both admins without having to copy/paste the file.

leonardr: This is a copy of a file in core -- we also have copies of log.py and util/http.py, possibly a couple others. There are a couple ways of avoiding this copy-paste. We could have a tiny 'util' submodule that only contained utilities like this, or we could split out the model portion of 'core' into its own submodule (used by circulation and metadata) and have 'core' just be stuff that doesn't depend on any particular model.

The amount of copy-and-pasted code is small enough that I'm not concerned about this, but I do think the model used by circulation and metadata is too monolithic and could be split up.

PR #138, opened 2019-09-10
Allow TOS to be stored as a small snippet of HTML in addition to a simple link.

leonardr:
This branch addresses https://jira.nypl.org/browse/SIMPLY-2277. The registry terms of service can be expressed as a simple link, as before, and/or as a snippet of explanatory HTML.

### DIFF
--- a/config.py
+++ b/config.py
@@ -39,9 +39,15 @@ class Configuration(object):
     ADOBE_VENDOR_ID_NODE_VALUE = "node_value"
     ADOBE_VENDOR_ID_DELEGATE_URL = "delegate_url"
 
-    # The URL to the document containing the terms of servier for
+    # The URL to the document containing the terms of service for
     # library registration
     REGISTRATION_TERMS_OF_SERVICE_URL = "registration_terms_of_service_url"
+
+    # An HTML snippet describing the terms of service for library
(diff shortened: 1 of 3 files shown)


config.py
```diff
@@ -39,10 +39,16 @@ class Configuration(object):
...
 
-    # The URL to the document containing the terms of servier for
+    # The URL to the document containing the terms of service for
     # library registration
     REGISTRATION_TERMS_OF_SERVICE_URL = "registration_terms_of_service_url"
 
+    # An HTML snippet describing the terms of service for library
+    # registration. It's better if this is a short snippet of text
+    # with a link, rather than the actual text of the terms of
+    # service.
```
adriana-alter: Does this presuppose that the admin setting up the library knows how to write HTML?  If so, it might be better to just add two fields to the library form--one for the text and one for the link--and let the front end interpolate everything.

leonardr: It supposes that the person setting up the library _registry_ knows how to write HTML, which I think is okay. If you don't know how to write HTML you can set the link instead and (once we implement this feature) let the circulation manager take care of the HTML.

adriana-alter: Ah, okay.  In that case, this branch looks good to me!

PR #140, opened 2019-10-03
Add admins table and login capability

adriana-alter:
Resolves https://jira.nypl.org/browse/SIMPLY-2273


leonardr (commented):
The data model looks but I'm not sure about the controller. It looks like you can create an admin account by trying to log in with a previously unseen username. Is this right?

I think it makes sense to allow that for the _first_ Admin -- the one you'd create right after setting up the site -- but it shouldn't be possible if there's already an Admin in the database. If there needs to be more than one Admin on a side, we can add to the administrative interface the ability to create new Admins with preset passwords.

leonardr (approved):

PR #142, opened 2019-11-01
Tell the front end how many Adobe IDs are associated with each library

adriana-alter:
Resolves https://jira.nypl.org/browse/SIMPLY-2348.  (Generating the data for each library, rather than running the query to list all the identifiers, turned out to be a lot easier for getting the information to the front end in a useful format.)

### DIFF
--- a/model.py
+++ b/model.py
@@ -419,15 +419,17 @@ class Library(Base):
     def pls_id(self):
         return ConfigurationSetting.for_library(Library.PLS_ID, self)
 
-    def number_of_patrons(self, _db):
+    @property
+    def number_of_patrons(self):
+        db = Session.object_session(self)
         # This is only meaningful if the library is in production.
         if not self.in_production:
-            return '0'
-        query = _db.query(DelegatedPatronIdentifier).filter(
+            return 0
+        query = db.query(DelegatedPatronIdentifier).filter(
             DelegatedPatronIdentifier.type==DelegatedPatronIdentifier.ADOBE_ACCOUNT_ID,
             DelegatedPatronIdentifier.library_id==self.id
         )
-        return str(query.count())
+        return query.count()
 
     @property
     def in_production(self):
--- a/model.py
+++ b/model.py
@@ -419,12 +419,15 @@ class Library(Base):
     def pls_id(self):
         return ConfigurationSetting.for_library(Library.PLS_ID, self)
 
-    @property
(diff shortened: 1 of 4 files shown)


model.py
```diff
@@ -419,6 +419,13 @@ def library_stage(self, value):
     def pls_id(self):
         return ConfigurationSetting.for_library(Library.PLS_ID, self)
 
+    @property
+    def number_of_patrons(self):
+        # This is only meaningful if the library is in production.
+        if not self.in_production:
+            return '0'
+        return str(len(filter(lambda id: (id.type == 'Adobe Account ID'), self.delegated_patron_identifiers)))
```
leonardr: I'm concerned this is going to fall over when there are 100,000 delegated patron identifiers (as there are now for NYPL and Open eBooks). If it's easier to get the count for one library than for all libraries at once, then that should be fine, but the counting really ought to be done inside the database. Otherwise we're creating 100,000 Python objects just to count how many there are.

adriana-alter: Ah.  That is...a very good point.  Can you clarify what you mean about doing the counting inside the database, though?

leonardr: If you create a query and call count() on it, you'll execute a SELECT COUNT statement and the database will return the number of rows rather than the actual rows.

There are ways to do even fancier queries with the `select()` method -- `Work.to_search_document` is an extreme example and `CoverageRecord.bulk_add` is a less extreme example. If you're interested I'm happy to go over this with you so we can figure out how to do this for every library in one query. But building a query with SQLAlchemy and calling `count()` on it will get us where we need to be in terms of performance.

model.py
```diff
@@ -419,6 +419,16 @@ def library_stage(self, value):
     def pls_id(self):
         return ConfigurationSetting.for_library(Library.PLS_ID, self)
 
+    def number_of_patrons(self, _db):
```
leonardr: You can get a database connection from inside a model object method by calling `Session.object_session(self)`, so there's no need to pass this in. See `Library.set_hyperlink` for an example.

model.py
```diff
@@ -419,6 +419,16 @@ def library_stage(self, value):
...
+        # This is only meaningful if the library is in production.
+        if not self.in_production:
+            return '0'
```
leonardr: This should return a number in both cases since the method is called "number". If it needs to be converted into a string that should happen outside.

PR #147, opened 2019-11-12
Filter out cancelled libraries in libraries_opds

adriana-alter:
Partially resolves https://jira.nypl.org/browse/SIMPLY-2349.  (I think the proposed changes to the admin endpoint can be mostly handled on the front end, so that will be a separate ticket and PR.)


adriana-alter:
I'll do it; the work is almost done anyway (it's a really easy fix!).

PR #159, opened 2020-03-06
Validate emails to register

adriana-alter:
This resolves https://jira.nypl.org/browse/SIMPLY-2640--the backend for https://jira.nypl.org/browse/SIMPLY-2605.  (This is just making it possible to validate the other emails from the interface--not requiring that all three emails are validated prior to registration, which will be covered later by https://jira.nypl.org/browse/SIMPLY-2606.)

### DIFF
--- a/controller.py
+++ b/controller.py
@@ -339,7 +339,9 @@ class LibraryRegistryController(BaseController):
             "help_email": Hyperlink.HELP_REL,
             "copyright_email": Hyperlink.COPYRIGHT_DESIGNATED_AGENT_REL
         }
-        hyperlink = Library.get_hyperlink(library, email_types[email])
+        hyperlink = None
+        if email_types[email]:
+            hyperlink = Library.get_hyperlink(library, email_types[email])
         if not hyperlink or not hyperlink.resource or isinstance(hyperlink, ProblemDetail):
             return INVALID_CONTACT_URI.detailed(
                 "The contact URI for this library is missing or invalid"
--- a/controller.py
+++ b/controller.py
@@ -340,7 +340,7 @@ class LibraryRegistryController(BaseController):
             "copyright_email": Hyperlink.COPYRIGHT_DESIGNATED_AGENT_REL
         }
         hyperlink = None
-        if email_types[email]:
+        if email_types.get(email):
             hyperlink = Library.get_hyperlink(library, email_types[email])
         if not hyperlink or not hyperlink.resource or isinstance(hyperlink, ProblemDetail):
             return INVALID_CONTACT_URI.detailed(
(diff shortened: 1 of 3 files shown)


controller.py
```diff
@@ -325,15 +330,20 @@ def _validated_at(self, hyperlink):
...
         library = self.library_for_request(uuid)
         if isinstance(library, ProblemDetail):
             return library
-        hyperlink = Library.get_hyperlink(library, Hyperlink.INTEGRATION_CONTACT_REL)
+        email_types = {
+            "contact_email": Hyperlink.INTEGRATION_CONTACT_REL,
+            "help_email": Hyperlink.HELP_REL,
+            "copyright_email": Hyperlink.COPYRIGHT_DESIGNATED_AGENT_REL
+        }
+        hyperlink = Library.get_hyperlink(library, email_types[email])
```
leonardr: I think you could crash this by passing in a junk value for `email` -- you'd then get a ValueError when the junk value wasn't in `email_types`. In this situation `hyperlink` should end up None and `get_hyperlink` probably shouldn't be called.

tests/test_controller.py
```diff
@@ -251,12 +251,12 @@ def _is_library(self, expected, actual, has_email=True):
                 actual_time = [actual_ts.year, actual_ts.month, actual_ts.day]
                 expected_time = [expected_ts.year, expected_ts.month, expected_ts.day]
                 eq_(actual_time, expected_time)
-            elif k == "contact_email":
+            elif "_email" in k:
```
leonardr: `k.endswith("_email")` will work the same way and avoid future false positives. This is in test code so it's not as important.

tests/test_controller.py
```diff
@@ -251,12 +251,12 @@ def _is_library(self, expected, actual, has_email=True):
...
                 if has_email:
-                    expected_contact_email = expected.name + "@library.org"
-                    eq_(flattened.get("contact_email"), expected_contact_email)
-            elif k == "validated":
-                eq_(flattened.get("validated"), "Not validated")
+                    expected_email = expected.name + "@library.org"
+                    eq_(flattened.get(k), expected_email)
+            elif "_validated" in k:
```
leonardr: Similarly here.

leonardr (changes requested):
This looks good with a minor change to make the form handling more robust.

controller.py
```diff
@@ -339,7 +339,9 @@ def validate_email(self):
             "help_email": Hyperlink.HELP_REL,
             "copyright_email": Hyperlink.COPYRIGHT_DESIGNATED_AGENT_REL
         }
-        hyperlink = Library.get_hyperlink(library, email_types[email])
+        hyperlink = None
+        if email_types[email]:
```
leonardr: This will still crash if `email` isn't a key in `email_types`. Try `email_types.get(email)` instead.

leonardr (approved):

PR #163, opened 2020-03-20
Send compressed library lists if the client supports it.

leonardr:
This branch adds compression support to the /libraries and /libraries/qa endpoints if the client requests them using the Accept-Encoding header. The library list compresses from 2.6 megs to 1.9 megs, a 25% improvement. According to @twaddington, the Android HTTP client sends `Accept-Encoding: gzip` by default, so we'll start seeing benefits immediately.

This fixes https://jira.nypl.org/browse/SIMPLY-1778. I talk about a lot of other stuff in that ticket, but it's now covered in other tickets.

### DIFF
--- a/app_helpers.py
+++ b/app_helpers.py
@@ -65,7 +65,6 @@ def compressible(f):
     def compressor(*args, **kwargs):
         @flask.after_this_request
         def compress(response):
-            set_trace()
             if (response.status_code < 200 or
                 response.status_code >= 300 or
                 'Content-Encoding' in response.headers):
--- a/app_helpers.py
+++ b/app_helpers.py
@@ -77,12 +77,13 @@ def compressible(f):
             if not 'gzip' in accept_encoding.lower():
                 return response
 
-            # At this point we know we'ree going to be changing the
+            # At this point we know we're going to be changing the
             # outgoing response.
 
             # TODO: I understand what direct_passthrough does, but am
-            # not sure what it has to do with this. This is pure
-            # copy-and-paste magic.
+            # not sure what it has to do with this, and commenting it
+            # out doesn't change the results or cause tests to
+            # fail. This is pure copy-and-paste magic.
             response.direct_passthrough = False
 
             buffer = BytesIO()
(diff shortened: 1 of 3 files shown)


app_helpers.py
```diff
@@ -44,3 +46,58 @@ def decorated(*args, **kwargs):
...
+            if (response.status_code < 200 or
+                response.status_code >= 300 or
+                'Content-Encoding' in response.headers):
+                # Don't encode anything other than a 2xx response
+                # code. Don't encode a response that's
+                # already been encoded.
+                return response
+
+            accept_encoding = flask.request.headers.get('Accept-Encoding', '')
+            if not 'gzip' in accept_encoding.lower():
```
twaddington: 👍

twaddington:
This looks good!

Unfortunately, while Android does add this by default, I took another look at our code and it seems like we override this header by setting it to `Accept-Encoding: identity` deep in our HTTP code.

https://github.com/NYPL-Simplified/Simplified-Android-Core/blob/develop/simplified-http-core/src/main/java/org/nypl/simplified/http/core/HTTP.java#L153

As a result, we won't see an immediate benefit by deploying this change.

I'm going try removing this line and doing some testing to make sure it still works (I believe removing it shouldn't harm anything). If it works fine, I can PR that along with the cache-control stuff.

> By default, this implementation of HttpURLConnection requests that servers use gzip compression and it automatically decompresses the data for callers of URLConnection.getInputStream(). &mdash; https://developer.android.com/reference/java/net/HttpURLConnection

leonardr:
The good news is the person who made this change back in 2015 -- @io7m -- is still on the project. Mark, what is your take on https://github.com/NYPL-Simplified/Simplified-Android-Core/commit/b14b87149e5923041acbcbfc81c2f68521a2ac47?

twaddington:
@leonardr yeah, I spoke with Mark. We both agree it seems counterproductive to disable gzip encoding globally like this. I'm gonna try removing it.

> If determining file size is the problem, we could do this through Transfer-Encoding instead.

I'm not sure what the impact of doing this would be offhand.

twaddington:
Good news is since you're checking the accept header this pull-request can be merged without harming anything on the Android side.

PR #164, opened 2020-03-20
Omit inline logo images from large feeds ('large' is configurable)

leonardr:
This branch adds a sitewide setting "large_feed_size", which is expected to be a number. At this size, a list of libraries is considered 'large' and is subject to different rules. Specifically, the OPDS entries generated for those libraries do not include the inline image containing the library's logo.

This addresses https://jira.nypl.org/browse/SIMPLY-2664 in a way that's easy to scale up, scale down, or remove in the future.


twaddington:
Great idea.

EdwinGuzman (approved):

PR #165, opened 2020-03-20
Add some tests.

leonardr:
This branch formalizes a hotfix I had to put into place on production as part of my work on https://jira.nypl.org/browse/SIMPLY-2664.

The code I had works fine if `libraries` is a list, but in real life, `libraries` is a SQLAlchemy query, and you get the size of a SQLAlchemy query with count(), not len()

I'm a little concerned about the performance effects once `libraries` becomes a much more complicated query involving geographic proximity, but there's also going to be a `LIMIT` attached to those queries, so we'll know how big they can get ahead of time. Also, once we're able to use those queries, there will effectively be no 'large' OPDS feeds and we can get rid of this whole thing.

### DIFF
--- a/tests/test_opds.py
+++ b/tests/test_opds.py
@@ -95,20 +95,27 @@ class TestOPDSCatalog(DatabaseTest):
         small_catalog = small_feed.catalog['catalogs']
         eq_([], small_catalog)
 
-    def _test_feed_is_large(self):
+    def test_feed_is_large(self):
         # Verify that the _feed_is_large helper method
         # works whether it's given a Python list or a SQLAlchemy query.
         setting = ConfigurationSetting.sitewide(
             self._db, Configuration.LARGE_FEED_SIZE
         )
         setting.value = 2
-        m = OPDSCatalog.feed_is_large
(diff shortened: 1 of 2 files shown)


tests/test_opds.py
```diff
@@ -88,6 +89,27 @@ def library_catalog(*args, **kwargs):
...
+
+    def _test_feed_is_large(self):
+        # Verify that the _feed_is_large helper method
+        # works whether it's given a Python list or a SQLAlchemy query.
+        setting = ConfigurationSetting.sitewide(
+            self._db, Configuration.LARGE_FEED_SIZE
+        )
+        setting.value = 2
+        m = OPDSCatalog.feed_is_large
+        list = [1,2,3]
```
EdwinGuzman: Shouldn't it be `_feed_is_large` with the leading underscore? Also, `list` isn't being used so how do the two tests below give different results?

leonardr: The problems you found are valid; I never finished writing this test. The tests were passing because this test wasn't being run; _its_ name starts with an underscore so nose didn't recognize it as a test.

PR #166, opened 2020-04-17
Promote nearby libraries to the top of the alphabetical list

leonardr:
This branch addresses https://jira.nypl.org/browse/SIMPLY-2703. The goal is to buy time for the client-side "onboarding" work by promoting libraries near a patron's current location (to help people in the common case), while still showing a big alphabetical library list (since there is no search functionality to help in the uncommon case).

Since different patrons now get different library feeds, I had to remove the caching code. While doing this I discovered that the caching code didn't actually do much, because I'd added it to the controller used by the admin interface (libraries), not the controller used by the mobile clients (libraries_opds). 

I also discovered some significant performance problems building the library feed, caused by having to go to the database to get the contact information for each library. I solved this using the same `joinedload` technique I've used elsewhere in this system -- this fetches related data model objects from the database at the same time as the `Library` objects we request.

These performance problems aren't nearly as bad in production as they are on my laptop (1.5 seconds to get a feed vs 20 seconds), possibly because the production server and the database are on the same network, but I was able to speed this up 4x locally by not going to the database, and we should see a proportional improvement in production.

### DIFF
--- a/app.py
+++ b/app.py
@@ -108,16 +108,16 @@ def confirm_resource(resource_id, secret):
 
 @app.route('/libraries')
 @compressible
-@has_location
+@uses_location
 @returns_problem_detail
-def libraries_opds():
+def libraries_opds(_location=None):
     return app.library_registry.registry_controller.libraries_opds(location=_location)
 
 @app.route('/libraries/qa')
 @compressible
-@has_location
+@uses_location
 @returns_problem_detail
-def libraries_qa():
+def libraries_qa(_location=None):
     return app.library_registry.registry_controller.libraries_opds(location=_location, live=False)
 
 @app.route('/admin/log_in', methods=["POST"])
--- a/app.py
+++ b/app.py
@@ -108,15 +108,17 @@ def confirm_resource(resource_id, secret):
 
 @app.route('/libraries')
 @compressible
+@has_location
 @returns_problem_detail
 def libraries_opds():
-    return app.library_registry.registry_controller.libraries_opds()
+    return app.library_registry.registry_controller.libraries_opds(location=_location)
 
 @app.route('/libraries/qa')
 @compressible
+@has_location
 @returns_problem_detail
 def libraries_qa():
-    return app.library_registry.registry_controller.libraries_opds(live=False)
+    return app.library_registry.registry_controller.libraries_opds(location=_location, live=False)
 
 @app.route('/admin/log_in', methods=["POST"])
 @returns_problem_detail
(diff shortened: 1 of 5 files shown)


app.py
```diff
@@ -108,15 +108,17 @@ def confirm_resource(resource_id, secret):
 
 @app.route('/libraries')
 @compressible
+@uses_location
```
leonardr, line 111: This is the same decorator the `/` route uses. It'll look in the `_location` query string for a latitude/longitude, and if that's not present it'll try IP geolocation.

opds.py
```diff
@@ -73,13 +73,17 @@ def __init__(self, _db, title, url, libraries, annotator=None,
 
         self.add_link_to_catalog(self.catalog, rel="self",
                                  href=url, type=self.OPDS_TYPE)
+        web_client_uri_template = ConfigurationSetting.sitewide(
+            _db, Configuration.WEB_CLIENT_URL
+        ).value
```
leonardr, line 78: Grabbing this out of the database before iterating over the list of libraries saves us one database query per library for the same data.

controller.py
```diff
@@ -173,7 +174,6 @@ def __init__(self, app, emailer_class=Emailer):
         super(LibraryRegistryController, self).__init__(app)
         self.annotator = LibraryRegistryAnnotator(app)
         self.log = self.app.log
-        self.library_list_cache = dict()
```
EdwinGuzman, line 176: This was the existing cache that was being used for the admin interface and not the mobile clients?

leonardr: Yeah. I meant it to be used by the mobile clients but I put it in the wrong method, and now it won't work anyway.

controller.py
```diff
@@ -266,16 +249,50 @@ def libraries(self, live=True):
...
+            # query, because it doesn't simply return Library objects,
+            # but it won't return more than five results.
+            nearby_libraries = Library.nearby(
+                self._db, location, production=live
+            ).limit(5).all()
+
+            # Exclude nearby libraries from the alphabetical query
+            # to get a list of faraway libraries.
+            faraway_libraries = alphabetical.filter(
+                ~Library.id.in_([x.id for x, distance in nearby_libraries])
```
EdwinGuzman, line 293: Are you excluding the nearby libraries by returning the inverse of the nearby libraries?

leonardr: I wouldn't put it like that. The `~Library.id.in_` syntax here gets translated to the SQL syntax `libraries.id NOT IN`.

EdwinGuzman: Ah, that's because this is still a SQL object, right?

leonardr: Right, every call to `filter` is adding another WHERE clause to the eventual query.

PR #171, opened 2021-03-23
Do a better job of avoiding referring to logos.

leonardr:
Early in the pandemic I wrote https://github.com/NYPL-Simplified/library_registry/pull/164 to eliminate library logos from the OPDS feed of all libraries. These logos were bloating the feed to the point that our increased number of clients couldn't reliably download it.

But I didn't eliminate the _server-side_ work of considering the logos; they're still loaded in from the database. And it turns out that when one field of a database table contains 500 kilobytes of data, queries involving that table can become mysteriously slow.

This branch expands on pull #164 to stop the `logo` field from being used in the "big list of libraries" query or extracted from the database unnecessarily. It also adds some logging to the process of feed generation so it'll be easier to track down future performance problems. And it freezes `requirements.txt` to the versions of dependencies currently deployed in production -- something that will soon be rendered moot by Dockerization.

This is a short-term fix; https://jira.nypl.org/browse/SIMPLY-3643 covers the long-term fix for this problem.


controller.py
```diff
@@ -272,33 +276,45 @@ def libraries_opds(self, live=True, location=None):
             joinedload('hyperlinks', 'resource'),
             joinedload('hyperlinks', 'resource', 'validation'),
         )
+        alphabetical = alphabetical.options(defer('logo'))
```
leonardr, line 279: This stops the database from pulling the `logo` field into memory. The presence of this huge field in memory while doing joins is apparently what causes the slowdown.

opds.py
```diff
@@ -156,7 +156,7 @@ def library_catalog(
                                     href=library.web_url,
                                     type="text/html")
 
-        if library.logo and include_logo:
+        if include_logo and library.logo:
```
leonardr, line 159: Without `include_logo` to short-circuit this code, we'll refer to `library.logo` for every library and fetch it from the database anyway. It will actually still be faster than retrieving it as part of the big query, but we don't need it at all right now.

nballenger (approved):
Looks great, tests all passed in Travis. Nice catch and fix.

Only comment is that for timing calls in the future it may be worth looking at whether any of the newer timing functions like [`time.thread_time()`](https://docs.python.org/3/library/time.html#time.thread_time), [`time.monotonic()`](https://docs.python.org/3/library/time.html#time.monotonic), or [`time.process_time()`](https://docs.python.org/3/library/time.html#time.process_time) are helpful, since `time.time()` isn't always accurate for sub-second timing, and returns epoch time rather than CPU time. (In this case I think it's totally fine since we're worried about 1s vs 10s.)

leonardr:
I went looking for a better way to time stuff, and `monotonic` looks great but it's exclusive to Python 3. We'll get there soon!

PR #176, opened 2021-04-13
Minimal changeset to move to Python 3 and a Docker install

nballenger:
Summary of changes:

* Update from Python 2.7 to 3.9
* Moves from `nose` to `pytest`, though without moving to use Pytest's fixture system
* Changes the two static routes (`registry-admin.css` and `registry-admin.js`) to be served via Nginx
* Adds Docker files and scripts to run locally and in production
* Dependency management is via Pipenv
* Documents install and usage in README
* Adds convenience commands for controlling the cluster via `make`


tests/__init__.py
```diff
@@ -11,8 +11,7 @@
 # Having problems with the database not being initialized? This module is
 # being imported twice through two different paths. Uncomment this
 # set_trace() and see where the second one is happening.
```
leonardr: Change the comment to something like 'put a breakpoint here'.

tests/test_controller.py
```diff
@@ -52,6 +48,10 @@
 from config import Configuration
 
 
+def eq_(a, b):
+    assert a == b
+
+
```
leonardr: You can take this out now.

tests/test_emailer.py
```diff
@@ -13,6 +10,9 @@
 )
 import quopri
 
+def eq_(a, b):
+    assert a == b
+
```
leonardr: Same here.

tests/test_geometry_loader.py
```diff
@@ -19,6 +16,10 @@
 from geometry_loader import GeometryLoader
 
 
+def eq_(a, b):
+    assert a == b
+
+
```
leonardr: I won't point all of these out but I'm pretty sure you can get rid of all of them.

tests/test_language.py
```diff
@@ -52,16 +53,16 @@ def test_string_to_alpha_3(self):
...
-        eq_(u"English/español", m(["eng", "spa"]))
-        eq_(u"español/English", m("spa,eng"))
-        eq_(u"español/English/Chinese", m(["spa","eng","chi"]))
-        assert_raises(ValueError, m, ["eng, nxx"])
+        eq_("español", m(['es']))
+        eq_("English/español", m(["eng", "spa"]))
+        eq_("español/English", m("spa,eng"))
+        eq_("español/English/Chinese", m(["spa","eng","chi"]))
+        with pytest.raises(ValueError):
+            m(["eng, nxx"])
```
leonardr, line 65: Well, there are a lot of these so I guess it can wait for a follow-up branch.

nballenger:
I went through and ditched all the `eq_` calls in favor of `assert` statements, and fixed that piece of in-line docs related to `set_trace()` .

nballenger:
Added info to the README about install/build and usage, and a `Makefile` with convenience commands. Also fixed the webapp entrypoint script, which was failing to actually wait for the database to become available before launching the app.

README.md
```diff
@@ -1,13 +1,51 @@
...
+
+### Running the Containers
+
+You can start up the local compose cluster in the background with:
+
+```shell
+make up
 ```
-CREATE EXTENSION postgis;
-CREATE EXTENSION fuzzystrmatch;
```
leonardr, line 6: I don't think you can remove this information from the human-readable documentation. In a non-development environment the database will be configured separately and this step will still need to be run manually, right? At least we need a pointer to the script that runs this for the Docker database image.

util/short_client_token.py
```diff
@@ -1,103 +1,100 @@
...
+            to_decode = to_decode.encode("utf8")
+
         to_decode = to_decode.replace(b":", b"+").replace(b";", b"/").replace(b"@", b"=")
-        return base64.decodestring(to_decode)
 
-    # The JWT spec takes January 1 1970 as the epoch.
-    JWT_EPOCH = datetime.datetime(1970, 1, 1)
+        return base64.decodebytes(to_decode)
+    
+    JWT_EPOCH = datetime.datetime(1970, 1, 1)   # The JWT spec takes January 1 1970 as the epoch.
```
leonardr, line 38: In circulation code we've changed all of our datetime handling so that our datetime objects are timezone-aware and UTC (rather than being timezone-naive). The fact that this appears to be local time is a problem, though it might be a problem we already have. Rather than just change this to a UTC timestamp (which might break the registry/CM relationship) let's investigate doing the timezone stuff in a follow-up branch.

leonardr:
I've been through the entire diff so far and it looks good overall. Apart from the current test failures, I pointed out two issues. The documentation issue should be addressed in this branch (or I need to understand why it's not an issue).

The question of whether the epoch is local time or UTC is more serious since a change could result in circulation managers starting to create Short Client Tokens that the registry considers invalid. (On the other hand, this could already be happening and a change would fix it -- that seems unlikely though.)

leonardr:
I filed https://jira.nypl.org/browse/SIMPLY-3705 to cover the datetime stuff.

README.md
nballenger: I just put non-Docker install instructions into the README, in b3bd736.

nballenger:
@leonardr - Agreed on a followup branch for the UTC stuff, since JWT tokens and timezones can both be pretty finicky. Not sure what was happening with those test failures--looks like a Travis hiccup where it wasn't propagating the database connection string environment variable and tried to default to an sqlite db. It corrected in the subsequent run where I added the README changes.

leonardr:
OK this is good to merge. But before you do, this is a convenient time to set up git flow the way we have on `circulation`:

* Rename `master` to `main`. `main` will stay on Python 2 until the Python 3 work is complete, in case we need to deploy hotfixes.
* Branch `develop` off of main and merge this branch into develop.

I did this for `circulation`, so let me know if you want to take care of it for this project as well.

nballenger:
@leonardr - It looks like I don't have permissions for the repo settings, so I can't change the default branch. I pushed `main` and `develop` branches though, so those are there now based on existing `master`.

leonardr:
I changed the default branch, so you should now be good to merge this into `develop`.

PR #181, opened 2021-04-22
Make it possible to obtain a human-readable description of a library's service area

leonardr:
This branch implements the core code necessary to fix https://jira.nypl.org/browse/SIMPLY-3456. If it's reasonable to associate a library with one specific place, the `Library.service_area_name` property will give you the name of that place.

All that remains is adding this information to the`metadata` section when generating a library's OPDS 2 entry. Since we need to keep two different OPDS 2 generators running in parallel (one for the huge list of 300 libraries, one for the smaller list that contains more information about fewer libraries), I added this to https://jira.nypl.org/browse/SIMPLY-3660 rather than putting the code in place as part of this branch.

### DIFF
--- a/model.py
+++ b/model.py
@@ -437,6 +437,42 @@ class Library(Base):
         prod = self.PRODUCTION_STAGE
         return self.library_stage == prod and self.registry_stage == prod
 
+    @property
+    def service_area_name(self):
+        """Describe the library's service area in a short string a human would
+        understand, e.g. "Kern County, CA".
+
+        TODO: We'll want to fetch a library's ServiceAreas (and their
+        Places) as part of the library query so that this doesn't result
+        in extra DB queries per library.
+
+        :return: A string, or None if the library's service area can't be
+           described as a short string.
+        """
+        # Group the service areas by type.
+        by_type = defaultdict(list)
+        for a in self.service_areas:
+            if a.place.type == Place.EVERYWHERE:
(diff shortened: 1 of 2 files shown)


model.py
```diff
@@ -437,6 +438,48 @@ def in_production(self):
...
+            if not a.place:
+                continue
+            if a.place.type == Place.EVERYWHERE:
+                # We already know that 'everywhere' won't work. Ignore
+                # it so it doesn't mask something more specific.
+                continue
+            by_type[a.type].add(a)
+
+        # If there is a single focus area, use it.
+        # Otherwise, if there is a single eligibility area, use that.
```
nballenger, line 469: I'll preface this by saying I don't know the data enough to actually know this is useful, so feel free to ignore. I wonder though if there's a potential low-lift win for cases where a library's service area is a list of zipcodes, where the service area name is expressed as `f"ZIP codes {', '.join(list_of_zips)}"`?

nballenger: Possibly even for any library where there are multiple service areas but they're all the same type, like `f"{service_area_type_plural}: {', '.join(service_area_names)}"`

leonardr: It's possible to generate a string like this, but the mockups don't show enough space for it, and I don't think seeing something that complicated would help anyone. I think the 'right' answer here is to make it possible for the library admin to set this to a freeform value like 'SF Bay Area'.

nballenger (approved):
Looks good to me.

PR #182, opened 2021-04-22
Treat an empty list of service areas as a lack of information

leonardr:
This fixes https://jira.nypl.org/browse/SIMPLY-3657. If an OPDS server doesn't list any service areas, we don't create any ServiceArea objects for it. Previously we were assuming such an OPDS server was a wide-open public server, but in practice such an OPDS server is a circulation manager that hasn't configured its libraries' service areas.

There's also a file, test_util_string_helpers.py, which has been sitting on my laptop for a while because I didn't add it after writing it. I discovered this mistake today when for the first time I ran the full library registry test suite under Python 3 and it gave me a couple of errors.

### DIFF
--- a/tests/test_scripts.py
+++ b/tests/test_scripts.py
@@ -9,6 +9,7 @@ from model import (
     ExternalIntegration,
     Library,
     Place,
+    ServiceArea,
     create,
     get_one,
 )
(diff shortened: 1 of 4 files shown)


tests/test_scripts.py
```diff
@@ -412,27 +413,36 @@ def test_success(self):
...
 
-        # Setting a service area with no focus area assigns that
+        # Setting an eligibility area with no focus area assigns that
         # service area to the library.
         args = ["--library=%s" % library.name,
                 '--service-area={"US": "everywhere"}']
         s.run(args)
         [area] = library.service_areas
         assert area.place == us
+        assert area.type == ServiceArea.FOCUS
```
leonardr, line 423: This is a little tricky because the command-line argument is `--service-area` but internally we call this an "eligibility area".

nballenger (approved):
Okay, paraphrasing my understanding of what's changing here:

* When an `AuthenticationDocument` object is created, its `service_area` and `focus_area` attributes are set via a call to `parse_service_area_and_focus_area()`
* In the case where no `service_area` value has been supplied to the `AuthenticationDocument` constructor, we previously had been setting the service area to cover everywhere
* This PR changes that behavior such that when no `service_area` is supplied to the constructor, the resulting `AuthenticationDocument` simply has an empty service area, rather than defaulting to everywhere.

Looks good to me, and the tests seem to be passing, so as long as my understanding checks out I think all is well.

PR #185, opened 2021-05-07
SIMPLY-3719: Make originating_ip() resilient to varied input

nballenger:
Previously a call to `flask_util.originating_ip()` would trigger an unhandled exception if the value in the `X-Forwarded-For` header contained more than a single IP address, which can happen when behind multiple layers of proxy server.

This makes `originating_ip()` more resilient to different values in `X-Forwarded-For` and the `request.remote_addr` property. It also explicitly restricts to publicly routable IP addresses, so we don't try to pass (for instance) a CIDR block IP into a geolocation-by-ip call.

I also had to fix some tests in `test_util_string_helpers.py` that were still using `nose`.

### DIFF
--- a/tests/test_flask_util.py
+++ b/tests/test_flask_util.py
@@ -1,4 +1,5 @@
 import ipaddress
+import re
 
 import pytest
 from flask import Flask, request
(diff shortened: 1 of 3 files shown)


leonardr (approved):
This looks good. Hopefully we won't need it much longer because we'll be removing the IP geolocation code, but I do think it's necessary to get a Python 3 Docker deploy to work.

PR #186, opened 2021-05-10
Provide service area information as part of a library's OPDS 2 entry

leonardr:
This branch adds information about a library's service area to its OPDS 2 entry. This means the _name_ of the service area (e.g. "New York, NY") and an OPDS 2 category based on the _type_ of place the library covers (e.g. "local" or "state"). Together, these changes take care of https://jira.nypl.org/browse/SIMPLY-3660.

Two additional notes:

* Like logo images, service area information is omitted for large feeds such as the huge ones at /libraries. This guarantees that any additional performance impact of looking up service area information isn't multiplied 300-fold, killing performance on SimplyE.
* I changed a couple of metadata keys to be compliant with the OPDS spec: 'updated' became 'modified' and 'distance' became 'schema:distance'. These aren't used by SimplyE AFAIK but I made the changes backwards-compatible just in case.

### DIFF
--- a/model.py
+++ b/model.py
@@ -211,6 +211,31 @@ def create(db, model, create_method='',
 
 Base = declarative_base()
 
+class LibraryType(object):
+    """Constant container for library types.
+
+    This is as defined here:
+
+    https://github.com/NYPL-Simplified/Simplified/wiki/LibraryRegistryPublicAPI#the-subject-scheme-for-library-types
+    """
+
+    SCHEME_URI = "http://librarysimplified.org/terms/library-types"
+    LOCAL = "local"
+    COUNTY = "county"
+    STATE = "state"
+    PROVINCE = "province"
+    NATIONAL = "national"
+    UNIVERSAL = "universal"
+
+    # Different nations use different terms for referring to their
+    # administrative divisions, which translates into different terms in
+    # the library type vocabulary.
+    ADMINISTRATIVE_DIVISION_TYPES = {
+        "US": STATE,
+        "CA" : PROVINCE,
+    }
+
+
 class Library(Base):
     """An entry in this table corresponds more or less to an OPDS server.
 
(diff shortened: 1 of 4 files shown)


model.py
```diff
@@ -471,14 +515,42 @@ def service_area_name(self):
...
+        this doesn't result in extra DB queries per library.
+
+        :return: A string, or None if the library's service area can't be
+           described as a short string.
+        """
+        service_area_place = self.service_area
+        if not service_area_place:
+            # There is no one place representing this library's service area.
+            return None
+        if service_area_place.type == Place.EVERYWHERE:
```
nballenger: May as well convert these to a single conditional, like

```Python
# If there's no one place that represents the service area, self.service_area is empty
# If the service area place is EVERYWHERE, that doesn't have a well-known name
if not service_area_place or service_area_place.type == Place.EVERYWHERE:
    return None
```

nballenger: Though looking at `human_friendly_name` there's already a check for `Place.EVERYWHERE` on [1407](https://github.com/NYPL-Simplified/library_registry/blob/5c3e9443dc99388300dbe8e161a480328288720b/model.py#L1407), so you could just have this be 

```Python
if self.service_area:
    return self.service_area.human_friendly_name
else:
    return None
```

model.py
```diff
@@ -1297,6 +1370,30 @@ def name_parts(cls, name):
...
+            # Whether this is a 'state' library, 'province' library,
+            # etc. depends on which nation it's in.
+            library_type = LibraryType.STATE
+            if self.parent and self.parent.type == Place.NATION:
+                library_type = LibraryType.ADMINISTRATIVE_DIVISION_TYPES.get(
+                    self.parent.abbreviated_name, library_type
+                )
+            return library_type
+        if self.type == Place.COUNTY:
+            return LibraryType.COUNTY
```
nballenger: May be worth converting these to a single `if ... elif ... else` stack, to save a couple cycles. My instinct would be to do that in descending order of most common type to least common, to keep it from entering the elifs in the majority of cases.

nballenger (approved):
Looks good. Couple minor things in-line, but nothing to worry about if you're strapped for time.

model.py
leonardr: `if..elif` is a good idea but I think the readability of going from "big" places at the top to "small" places at the bottom beats trying to figure out which types of libraries will be most common.

PR #191, opened 2021-05-26
Get Adobe endpoints to use utf8 decoded input

nballenger:
As per [SIMPLY-3744](https://jira.nypl.org/browse/SIMPLY-3744), the py3 registry in QA was failing to successfully complete the Adobe sign in loop, and was returning the following error:

```xml
<error xmlns="http://ns.adobe.com/adept" data="E_NYPL_AUTH 'bytes' object has no attribute 'xpath'"/>
```

This changeset makes the Adobe endpoints for sign in and user info read in from `flask.request.data.decode('utf8')` instead of just `flask.request.data`, which is a bytestring (and not able to be parsed by the adobe request parser).

Unfortunately our tests in `test_adobe_vendor_id.py` would not have caught this, because they don't engage the actual flask request/response cycle. I'm not reconfiguring to add those tests in this PR because it would slow things down quite a lot, but I will add a TODO item to make sure they're testable in future.

### DIFF
--- a/adobe_vendor_id.py
+++ b/adobe_vendor_id.py
@@ -32,7 +32,7 @@ class AdobeVendorIDController:
         """Process an incoming signInRequest document."""
         __transaction = self._db.begin_nested()
         output = self.request_handler.handle_signin_request(
-            request.data,
+            request.data.decode('utf8'),
             self.model.standard_lookup,
             self.model.authdata_lookup
         )
@@ -43,7 +43,7 @@ class AdobeVendorIDController:
     def userinfo_handler(self):
         """Process an incoming userInfoRequest document."""
         output = self.request_handler.handle_accountinfo_request(
-            request.data,
+            request.data.decode('utf8'),
             self.model.urn_to_label
         )
         return Response(output, 200, {"Content-Type": "application/xml"})
--- a/adobe_vendor_id.py
+++ b/adobe_vendor_id.py
@@ -1,16 +1,15 @@
-import flask
-from flask import Response
-
 import re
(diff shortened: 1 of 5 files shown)


adobe_vendor_id.py
```diff
@@ -36,15 +32,20 @@ def signin_handler(self):
         """Process an incoming signInRequest document."""
         __transaction = self._db.begin_nested()
         output = self.request_handler.handle_signin_request(
-            flask.request.data, self.model.standard_lookup,
-            self.model.authdata_lookup)
+            request.data.decode('utf8'),
```
nballenger, line 35: This is one of the non-formatting changes.

adobe_vendor_id.py
```diff
@@ -36,15 +32,20 @@ def signin_handler(self):
...
+        )
         __transaction.commit()
+
         return Response(output, 200, {"Content-Type": "application/xml"})
 
     def userinfo_handler(self):
         """Process an incoming userInfoRequest document."""
         output = self.request_handler.handle_accountinfo_request(
-            flask.request.data, self.model.urn_to_label)
+            request.data.decode('utf8'),
```
nballenger, line 46: This is the other non-formatting change.

nballenger:
Though I didn't add unit tests for this change, I did build and deploy an image with these changes to the ECS service backing qa-libraryregistry.librarysimplified.org. I then manually confirmed that it was able to correctly parse a short token, via the updated script `adobe-id-for-short-client-token` in the circulation repo.

PR #196, opened 2021-06-28
Update build deploy nomenclature

nballenger:
Changing some labeling for clarity around build and deploy.


nballenger:
I merged this to `develop` already. Merging to the deploy branches so I don't have to merge develop to QA again.

PR #199, opened 2021-06-30
Optimize the generation of the huge library lists used by the admin interface

leonardr:
This branch optimizes the `libraries` controller method used by the admin interface to build a detailed list of information for all libraries in the system. It should go some way towards addressing https://jira.nypl.org/browse/SIMPLY-3811, but I don't know whether it will address all the performance problems described in that ticket. My goal was just to get response time fast enough that we can finish our Python 3 acceptance testing.

I made two types of changes:

1. I introduced a new method, `Library.patron_counts_by_library`, which uses a single database query to retrieve patron counts for a list of libraries, rather than doing it one library at a time. The `libraries` method calls this method on all the libraries it's dealing with, and passes the appropriate patron count into `library_details`. If someone calls `library_details` without including this information, the fallback behavior is to go back to calculating this value for an individual library.
2. I used the same `joinedload` trick used in `libraries_opds` to pull all the ORM objects necessary to generate the `libraries` data in a single query. This also required changing some code and helper methods to make use of the preloaded objects instead of fetching those same objects from the database.


controller.py
```diff
@@ -236,14 +236,38 @@ def libraries(self, live=True):
...
+        alphabetical = self._db.query(Library).order_by(Library.name)
+
+        # Load all the ORM objects we'll need for these libraries in a single query.
+        alphabetical = alphabetical.options(
+            joinedload(Library.hyperlinks),
+            joinedload('hyperlinks', 'resource'),
+            joinedload('hyperlinks', 'resource', 'validation'),
+            joinedload(Library.service_areas),
+            joinedload('service_areas', 'place'),
+            joinedload('service_areas', 'place', 'parent'),
```
leonardr, line 248: We often need the parent of the library's service area when generating its human-readable name (e.g. the "IL" in "Cook County, IL").

model.py
```diff
@@ -462,6 +467,42 @@ def number_of_patrons(self):
...
+            DelegatedPatronIdentifier
+        )
+        rows = _db.execute(counts)
+
+        # Convert the results to a dictionary.
+        results = dict()
+        for (library_id, count) in rows:
+            results[library_id] = count
+
+        return results
```
nballenger, line 504: For a list of 2-tuples turns out you can just call `dict(list_of_2_tuples)`, which I didn't realize till just now. So you could just do `return dict(rows)` here.

There's a small efficiency difference to doing it that way, though it's unlikely to matter at the scale we're working at. This script:

```Python
import random, timeit, uuid

rows = [(str(uuid.uuid4()), random.randint(1000, 100000)) for _ in range(1, 1001)]

def by_for_loop():
  results = {}
  for (library_id, count) in rows:
    results[library_id] = count
  return results

def by_dict_call():
  return dict(rows)

loop_time = timeit.timeit(by_for_loop, number=10000)

dict_time = timeit.timeit(by_dict_call, number=10000)

print("For loop time: ", loop_time)
print("dict() time:   ", dict_time)
```

outputs:

```
For loop time:  0.388839666
dict() time:    0.28115554200000004
```

nballenger (approved):
Looks good, one inline comment about `dict(list_of_2_tuples)` vs. a for loop.

model.py
leonardr: That's good to know.  In this case, though, dict(rows) doesn't work because `rows` is a result set, not an actual list. Converting it to a list requires iterating over it, so I'm not convinced there will be a benefit here. I'll keep this in mind in general though.

PR #200, opened 2021-06-30
SIMPLY-3821, unconfigured/unresponsive mailer can cause uncaught ISE

nballenger:
If a registry instance does not have a configured/working mail integration, a call to register a library will fail with an uncaught exception. This PR addresses that by adding a ProblemDetail called `UNABLE_TO_NOTIFY`, which is sent if a registration is attempted when no mailer integration is working.

Note: There are a bunch of syntactical / readability / flake8 changes in the first three commits. Everything that's actually a functional change is in fadd3a2. I've made line-item notes on the real changes.


config.py
```diff
@@ -25,6 +25,10 @@ class CannotLoadConfiguration(Exception):
     pass
 
 
+class CannotSendEmail(Exception):
+    pass
```
nballenger, line 29: New exception type, raised when an `Emailer` instance's `send()` method gets an exception when it tries calling `_send_email()`.

controller.py
```diff
@@ -708,12 +700,16 @@ def register(self, do_get=HTTP.debuggable_get):
...
+                except SMTPException:
+                    # We were unable to send the email due to an SMTP error
                     return INTEGRATION_ERROR.detailed(
                         _("SMTP error while sending email to %(address)s",
                           address=hyperlink.resource.href)
                     )
+                except CannotSendEmail:
+                    return UNABLE_TO_NOTIFY.detailed(
+                        _("The Registry was unable to send a notification email.")
+                    )
```
nballenger, line 712: If the `Emailer` instance raised `CannotSendEmail` then the email integration is broken in a wider-ranging way than would raise an `SMTPException`, so we're going to abort the registration by returning an `UNABLE_TO_NOTIFY` document.

emailer.py
```diff
@@ -70,6 +76,73 @@ class Emailer(object):
...
+    def __init__(self, smtp_username, smtp_password, smtp_host, smtp_port, from_name, from_address, templates):
+        config_errors = []
+        required_parameters = ('smtp_username', 'smtp_password', 'smtp_host', 'smtp_port', 'from_name', 'from_address')
+        for param_name in required_parameters:
+            if not locals()[param_name]:
+                config_errors.append(param_name)
+
+        if config_errors:
+            msg = "Emailer instantiated with missing params: " + ", ".join(config_errors)
+            raise CannotLoadConfiguration(msg)
```
nballenger, line 89: Change unrelated to SIMPLY-3821, but of minor note: I switched out the individual `if` statements for this, to raise a more comprehensive error message.

emailer.py
```diff
@@ -70,6 +76,73 @@ class Emailer(object):
...
+        template = self.templates[email_type]
+        from_header = '%s <%s>' % (self.from_name, self.from_address)
+        kwargs['from_address'] = self.from_address
+        kwargs['to_address'] = to_address
+        body = template.body(from_header, to_address, **kwargs)
+
+        try:
+            self._send_email(to_address, body, smtp)
+        except Exception as exc:
+            raise CannotSendEmail(exc)
```
nballenger, line 131: Attempt to actually send the email. If a generic exception comes through, raise `CannotSendEmail`.

problem_details.py
```diff
@@ -1,72 +1,78 @@
...
     401,
-    title=_("The username or password is incorrect.")
+    title=lgt("The username or password is incorrect.")
+)
+
+UNABLE_TO_NOTIFY = pd(
+    "http://librarysimplified.org/terms/problem/unable-to-notify",
+    500,
+    title=lgt("Registry server unable to send notification emails.")
 )
```
nballenger, line 78: New `ProblemDetail` instance to return.

tests/test_controller.py
```diff
@@ -1430,6 +1440,51 @@ def send(self, *args, **kwargs):
         assert response.uri == INTEGRATION_ERROR.uri
         assert response.detail == "SMTP error while sending email to mailto:help@library.org"
 
+    def test_registration_fails_if_email_server_unusable(self):
```
nballenger, line 1443: Added a test for the missing or unresponsive mail server, as distinct from one raising an SMTP problem.

tests/test_emailer.py
```diff
@@ -302,16 +282,28 @@ def test_send(self):
...
             "To: you@library",
-            "subject Value".replace(" ", "_"), # Part of the encoding process.
+            "subject Value".replace(" ", "_"),  # Part of the encoding process.
             "Hello, you@library, this is me@registry."
         ]:
             print(phrase)
             assert phrase in body
         assert smtp == mock_smtp
 
+    def test_send_failure(self):
```
nballenger, line 292: Tests for send failure raising the new exception at the `Emailer` level.

leonardr (approved):

PR #202, opened 2021-07-26
Simply 3857/add static files

oliviawongnyc:
**Change**
- Added back in static file routes.
- Added in-line comments about the static file routes.

**Reason**
- This change enables the app to be run locally without Docker.

**Tests**
- With this re-introduced code, I confirmed the app still runs locally with and without Docker.


app.py
```diff
@@ -251,9 +271,9 @@ def admin_view():
         url = sys.argv[1]
     else:
         url = ConfigurationSetting.sitewide(_db, Configuration.BASE_URL).value
-    url = url or 'http://localhost:7000/'
-    scheme, netloc, path, parameters, query, fragment = urllib.parse.urlparse(
-        url)
+        url = url or 'http://localhost:7000/'
+        scheme, netloc, path, parameters, query, fragment = urllib.parse.urlparse(
+            url)
```
leonardr: I don't get why this no longer happens when `url` is provided on the command-line.

oliviawongnyc: Are you saying that this code is supposed to enable someone to run the app locally without Docker?

leonardr: Yes, the `__name__ == __main__` bit is run when someone just runs `python app.py` locally. There are three possibilities: 

1. You specify the base URL on the command line (`python app.py http://localhost:2301/`)
2. You don't specify a base URL (`python app.py`) and there is a BASE_URL set in the database configuration
3. You don't specify a base URL (`python app.py`) and there is no BASE_URL set, so the default of http://localhost:7000/ is used

In all three cases we need to set `netloc` for the piece of code immediately following this code. By changing the indentation you've set it up so that `netloc` is only set in case 2, and case 3 can't happen at all. (Update: actually I think case 3 still happens, just through a different route.)

I don't think you need to change this code at all; you can leave it as is.

leonardr (approved):

app.py
oliviawongnyc: Thanks for approving, Leonard! Fixed the indentation, as you saw. With the current set up – if you're running the app locally without Docker – you must execute `FLASK_APP=app.py pipenv run flask run` and navigate to http://localhost:5000/admin. `python app.py` will not work.

leonardr: That's fine, I think I just have an old pre-pipenv setup on my laptop.

oliviawongnyc: I see! Okay! Thank you.

PR #204, opened 2021-08-03
Updates registry_admin version number + CHANGELOG

oliviawongnyc:
**Change**
- Updated registry_admin version number from v1.4.14 to v1.4.15.

**Reason**
- Updated the SearchForm component in the registry_admin to display a loading message before the search is completed. Need to update the version in the library_registry to deploy this new feature to QA.

### DIFF
--- a/package.json
+++ b/package.json
@@ -10,7 +10,7 @@
   "author": "NYPL",
   "license": "Apache-2.0",
   "dependencies": {
-    "simplified-registry-admin": "1.4.14"
+    "simplified-registry-admin": "1.4.15"
   },
   "homepage": "https://github.com/NYPL-Simplified/library_registry#readme"
 }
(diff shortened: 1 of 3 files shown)


package.json
```diff
@@ -10,7 +10,7 @@
   "author": "NYPL",
   "license": "Apache-2.0",
   "dependencies": {
-    "simplified-registry-admin": "1.4.14"
```
EdwinGuzman, line 13: Update the version above as well.

oliviawongnyc: I think it is updated? See below!

EdwinGuzman: The registry admin package was updated but this app should be updated as well. In the changelog you have the new version as 1.0.11 so it should be updated on line 3 as well.

oliviawongnyc: Oh shoot!

EdwinGuzman (approved):

PR #210, opened 2021-08-23
Merging the current QA branch to prod for deploy

nballenger:
@leonardr - There's more in here than I would have thought. Do we need a round of testing in QA before we go to prod with this change set?


leonardr:
I'm not sure why all those _commits_ are showing up, but the number of _changes_ is relatively small. It basically looks like the PRs from the end of June, and each of them have been tested on QA.

That's enough of a change that I don't want to deploy at the end of the day, but if you want to run the basic integration tests we came up with for the Python 3 migration I'm OK with deploying this tomorrow morning.

PR #215, opened 2021-10-21
Refactor tests

leonardr:
This PR represents Nick's big library registry refactoring project.


library_registry/log.py
```diff
@@ -217,7 +216,7 @@ def loggly_handler(cls, externalintegration):
         try:
             url = cls._interpolate_loggly_url(url, token)
         except (TypeError, KeyError) as e:
-            raise CannotLoadConfiguraiton(
+            raise CannotLoadConfiguration(
```
leonardr: Whoa.

leonardr: This code was originally copied from core, so I checked there to verify it's been fixed there.

tests/conftest.py
```diff
@@ -0,0 +1,631 @@
...
+
+@pytest.fixture
+def connecticut_state(db_session, create_test_place, crude_us):
+    place = create_test_place(
+        db_session, external_id="09", external_name="Connecticut", place_type=Place.STATE,
+        abbreviated_name="CT", parent=crude_us,
+        geometry=(TEST_DATA_DIR / 'ct_state_geojson.json').read_text()
+    )
+    db_session.commit()
+    yield place
```
leonardr, line 427: Some of these take a long time to stick into the database (because the GeoJSON is complex), so you might think about scoping them to the test session.

library_registry/util/short_client_token.py
```diff
@@ -98,3 +127,159 @@ def _encode(self, library_short_name, library_secret, patron_identifier, expires
...
+        if not account_id:  # The delegates couldn't help us; let's try to do it ourselves.
+            try:
+                signature = self.adobe_base64_decode(password)
+            except Exception:
+                raise ValueError("Invalid password: %s" % password)
+
+            (patron_identifier, account_id) = self._decode(_db, username, signature)
+
+        # If we got this far, we have a Library, a patron_identifier, and an account_id.
+        from library_registry.model import DelegatedPatronIdentifier
```
leonardr, line 212: Since this plays with model objects, it might make sense to put it with the rest of the model utilities. Either way, I think   ShortClientTokenTool and ShortClientTokenDecoder should remain in the same file.

tests/test_scripts.py
```diff
@@ -29,109 +29,119 @@
...
 
-        # If no library is identified by name, the output of
-        # all_libraries is used as the list of libraries.
+        # If no library is identified by name, the output of all_libraries is used as the list of libraries.
         assert script.libraries() == script.all_libraries_return_value
 
-    def test_all_libraries(self):
+        db_session.delete(library)
+        db_session.delete(ignored)
+        db_session.commit()
```
leonardr: Presumably this has always been necessary in this code base and I never noticed before, but I think we need to introduce the code from core which runs each individual test in a database transaction that gets rolled back at the end. I don't like having to add this kind of cleanup code to the end of the tests.

leonardr:
This looks good overall. I think you uncovered a preexisting problem around database objects not being cleaned up between tests. As I mentioned earlier, I think it's better to resolve that by running each test in a database transaction that gets rolled back. That can also accommodate creating certain complex database objects (e.g. geographic places) ahead of time and tearing them down once the test is complete.

tests/test_scripts.py
nballenger: The teardown stuff has been bothering me too, but I'm not sure if it can all be handled by a transaction rollback in the session fixture. I need to go look at the sqlalchemy docs more, but I think it's the case that once you get a session handle, you can issue a commit on it, and then any rollback would only get rid of anything that had been added to the session after that commit. If there's a way to have a session kill everything that happened on it, that'd be useful--maybe mock the commit somehow? I'll take a look at what's in core and see how it handles it.

tests/conftest.py
nballenger: Good call, I'll give it a shot.

library_registry/log.py
nballenger: I don't know if it just never reaches that codepath in practice or we've been having uncaught exceptions.

tests/test_scripts.py
leonardr: This is going on memory, but in core and circulation we run the entire test suite in a transaction, and then run each individual test in a subtransaction. That way you can commit inside a test but the whole subtransaction will be rolled back at the end of the test.

leonardr:
I've reviewed the new changes and they also look fine. I'd like to merge this now and handle the "run each test in its own subtransaction" thing later since this refactoring isn't our main priority right now.

nballenger:
I'm not worrying about the transaction stuff, that's definitely an optimization for later. I had to get the last file of tests switched over. Unfortunately that was `test_controller.py`, which has a bunch of pretty complicated functional tests. That's done though, so if you want to take a look at that last commit (52f9f9d) I think we can merge.

leonardr:
OK, yes, everything looks good.

nballenger (approved):

PR #216, opened 2021-11-15
Updates to the container setup

nballenger:
Things this does:

* Explicitly runs the local db on port 5433, to avoid port collisions with other postgres instances
* Removes the webpack watch job in favor of building static front end files during the image build
* Renames some of the volumes and services in the compose file for clarity
* Updates the gunicorn config file to account for having placed the app in a package

If you're reviewing, it'd be awesome if you could pull the branch and do:

1. `make build`
2. `make up`
3. `make test`

Hopefully that runs the tests without issue, on either arm64 or amd64 chips.

After that, and before you do `make down`, if you could load the admin in a browser at `http://localhost/admin/` and click around a bit, that'd be super.


docker/gunicorn.conf.py
```diff
@@ -8,7 +8,7 @@
                if d.startswith("simplye_app-") and os.path.isdir(VENV_BASE_DIR / d)][0]
 
 # Shared Settings
-wsgi_app = "app:app"
+wsgi_app = "library_registry.app:app"
```
nballenger, line 11: @keithbauer was having issues with the Gunicorn process dying on start, because it wasn't finding the app object. Turns out I didn't update the path in `wsgi_app` to reflect making the app a package. Whoops!

nballenger:
Added some revised documentation in 064d018. Still need to fill out the Deployment.md file though.

leonardr:
`make test` fails for me because it's still looking for the database on port 5432.

Dockerfile
```diff
@@ -153,6 +154,23 @@ RUN set -ex \
...
  && apk del --no-network .build-deps
 
+# Build a static version of the front end to serve
+COPY ./package*.json ./
+
+RUN set -ex \
+ && apk add --no-cache --virtual .node-build-deps \
+    make \
+    build-base \
+    python2 \
```
leonardr, line 163: Who's using Python 2?

nballenger: It's a build requirement of `node-gyp` at whatever version we're using (or whatever version one of our dependencies is using, it may just get built for `node-sass`). We should really update those front end dependencies.

leonardr:
This may be my actual `make test` problem, an error initializing the database when I run `make up-watch`:

```
registry_db        | initdb: error: could not create directory "/var/lib/postgresql/data/pg_wal": No space left on device
registry_db        | initdb: removing contents of data directory "/var/lib/postgresql/data"
```

Presumably the device on which there's no space is _inside_ the Docker container. I don't know enough to diagnose this but I can answer any questions.

leonardr:
Nick helped me allocate more space on my hard drive for Docker images. Now `make up` works and `make db-session` works, but `make test` still gives me the error indicating it's trying to connect to port 5432.

nballenger:
That could be caused by having values for the database strings in a `.env` file--happened to me for a bit, and was super confusing. Docker compose will read `.env` and use the contents to override environment values set in the compose file or docker file.

leonardr:
Again, that's exactly the issue. Commenting out my .env file makes `make test` run to completion on my non-M1 Mac.

leonardr (approved):
I've now run the tests and looked at a running instance in my web browser--everything looks good.

nballenger:
I pushed changes in 9d30dc7 that rename all the instances of `/simplye_` to `/simplified_`. I also removed the `TESTING=1` environment variable from the local webapp build, and supplied it to the test commands in the Makefile instead. That's so that using the webapp doesn't write to the test database.

You have to rebuild to get those changes, but I've run the test suite against new images locally and passed everything.

I still need to rebuild the `active` image and run tests on that, since that's how we get our production images. I'll also need to review the Bamboo build job to make sure nothing needs to change there.

nballenger:
I've built and tested the `active` image, everything passes.

leonardr (approved):

leonardr (approved):

PR #217, opened 2021-11-19
Rollback test transactions

keithbauer:
This adds transaction rollback to `db_session`.

### DIFF
--- a/library_registry/app.py
+++ b/library_registry/app.py
@@ -27,13 +27,17 @@ db_url = Configuration.database_url(test=TESTING)
 test_db_url = Configuration.database_url(test=True)
 
 
-def create_app(testing=False):
+def create_app(testing=False, db_session_obj=None):
     app = Flask(__name__)
     babel.init_app(app)
 
-    SessionManager.initialize(db_url)
-    session_factory = SessionManager.sessionmaker(db_url)
-    _db = flask_scoped_session(session_factory, app)
+
+    if testing and db_session_obj:
+        _db = db_session_obj
+    else:
+        SessionManager.initialize(db_url)
+        session_factory = SessionManager.sessionmaker(db_url)
+        _db = flask_scoped_session(session_factory, app)
 
     log_level = LogConfiguration.initialize(_db, testing=testing)
     debug = log_level == 'DEBUG'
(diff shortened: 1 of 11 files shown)


nballenger:
So I actually think addressing the session mismatch issues in the `db_session` fixture won't be sufficient. The other failing test made me realize that the app's db session is going to get set as an attribute on all or most of the controller class instances that get set up during `LibraryRegistry.setup_controllers()`. If we don't definitively set it prior to that in `app.create_app()`, we'd have to replace it everywhere it had been set as an attribute of a downstream object. If not, those references to the original session will persist in the controller objects, even after the app object's `_db` points to some new session.

I think the fix is to alter `app.create_app()` to accept a db session object parameter, which can only get used during testing. Something like this:

```Python
def create_app(testing=False, db_session_obj=None):
    app = Flask(__name__)
    babel.init_app(app)

    if testing and db_session_obj:
        _db = db_session_obj
    else:
        SessionManager.initialize(db_url)
        session_factory = SessionManager.sessionmaker(db_url)
        _db = flask_scoped_session(session_factory, app)        

    log_level = LogConfiguration.initialize(_db, testing=testing)
    debug = log_level == 'DEBUG'
    app.config['DEBUG'] = debug
    app.debug = debug
    app._db = _db
    [...]
```

That will also mean that the `app` fixture has to become function scoped instead of session scoped, since it can't be created until after the `db_session` fixture is created. Which means shuffling the things in conftest to look like:

```Python
@pytest.fixture
def db_engine():
    engine = create_engine(test_db_url)
    yield engine
    engine.dispose()

@pytest.fixture
def db_session(db_engine):
    with db_engine.connect() as connection:
        transaction = connection.begin_nested()
        session = Session(connection)
        yield session
        transaction.rollback()
        session.close()

@pytest.fixture
def app(db_session):
    app = create_app(testing=True, db_session_obj=db_session)
    app.secret_key = "SUPER SECRET TESTING SECRET"
    yield app

@pytest.fixture
def client(app):
    with app.test_client() as client:
        yield client
```

nballenger:
FWIW I made those changes locally, and all tests pass in the same amount of time as the runs had been taking before.

leonardr:
I like having the database connection be the ancestor of all the other fixtures. I also prefer having a separate app for each test since that lets us mess around with the app. My only concern would be performance (I think we'd get a big performance hit if we tried this in the circulation manager right now), but if there's no performance hit I think this is perfect.

leonardr (approved):
I think this looks great now.

nballenger (approved):
Looks good! Nicely done.

PR #221, opened 2022-03-11
Adds new syntax to docker compose and make commands

oliviawongnyc:
- Adds syntax to the docker-compose file that tells docker to look for env variables for databases before falling back to the default.
- Adds new make commands which run the webapp container without the database container.


nballenger (approved):
👍

PR #225, opened 2022-04-07
Merge Flask blueprints

nyplwesclark:
## Description

This work is for ticket [SIMPLY-4045](https://jira.nypl.org/browse/SIMPLY-4045).

This PR:

- Creates a new file structure of groupings called admin, DRM, and library protocol with logically-related functionality.
- Uses Flask 2.0 Blueprint feature to modularize routes and views
- Moved AdobeVendorIDController to drm/controller.py
- Refactors part of LibraryRegistryController in library_registry/controller.py to AdminController in library_registry/admin/controller.py
- Adds AdminController to setup_controllers of LibraryRegistry in library_registry/controller.py
- Adds test_blueprint_routes.py to tests/ to test specific endpoints that are a part of a new Blueprint and updates test_controller.py in tests/ to work with Blueprints and test the new AdminController.

This PR specifically does not:
- Refactor code or make major changes to library protocol functionality.

## Motivation and Context

These changes are part of general code improvement and infrastructure upgrade for the library registry. Using Flask 2.0 and the Flask Blueprint feature and philosophy, the library registry is logically grouped into different components. Because the code is grouped logically and refactored into common and shared patterns, the library registry codebase is simplified, which should improve code readability and maintainability as well as overall speed of development. For example, it should be more intuitive that administrative endpoints are located in admin/routes.py instead of in their previous location, app.py, and it should be more intuitive that code for administrative functionality and development is located in admin/.

## How Has This Been Tested?

Set up a local environment following these [local development instructions](https://github.com/NYPL-Simplified/library_registry/blob/develop/docs/Development.md).

Testing used the **Makefile** in the root of the repository and the **make test** command. All tests passed locally.

## Checklist:

- [ ] I have updated the documentation accordingly.
- [x] All new and existing tests passed.


leonardr (changes requested):
This is a good first branch. We talked about the next steps to complete this work on the registry so that we're confident in moving on to the circulation manager: 

* Split the blueprint that manages the public-facing list of libraries from the blueprint that manages the library registration protocol.
* Move the rest of the controller code into the appropriate blueprints, the way this branch does with the admin code.
* Figure out testing???

All those question marks are because I still don't know how the blueprint routing code should be tested. End-to-end tests like the ones you put in TestAdminBlueprintRoutes are useful for quickly demonstrating what a successful request looks like, but as soon as things get complicated, it's simpler to test the routing code ("does a request to /aaa/bbb call the right controller method with the right arguments?") separately from the controller itself.

For the circulation manager we developed [a pretty complex set of fixtures](https://github.com/NYPL-Simplified/circulation/blob/develop/tests/test_routes.py) that let us separate those two types of tests, and I was hoping we'd be able to do without those fixtures by moving to blueprints.

Take out the stub tests for now, and once you're happy with the blueprint code, let's take a look at it (and the Flask docs) and see if there's a way to test the routes without all the fixtures we added to the circ manager. (But we should add the fixtures rather than do incomplete or inefficient tests of the routing code.)

leonardr (approved):
I think this is ready to merge, unless you'd rather merge in all the follow-ups and put the whole thing into `develop` at once.

nyplwesclark:
Great. We'll go ahead and merge. Thank you!

PR #228, opened 2022-04-27
Erd 26 security fix add authentication check to admin apis

nyplwesclark:
## Description

This work is for bug [ERD-26](https://jira.nypl.org/browse/ERD-26).

This PR:

- Creates a decorator in library_registry/admin/decorators.py to check if a user has been assigned a session cookie. If the user has not, then INVALID_CREDENTIALS (from library_registry/problem_details.py) is returned, which is a 401 with an incorrect credentials message. 
- Decorates the relevant admin routes to require that a user be authenticated and, if not, receive a 401 status code.

## Motivation and Context

Most admin routes did not check for authentication. This was not entirely terrible since any accessible information was already public, and it did not appear that the secret between the registry and circulation manager could be obtained. However, this still isn't a good practice, and it is important to shore this up, particularly as we move forward with improving authentication in general.

## How Has This Been Tested?

Set up a local environment following these [local development instructions](https://github.com/NYPL-Simplified/library_registry/blob/develop/docs/Development.md).

Test using the **Makefile** in the root of the repository and the **make test** command. All tests passed locally. Specifically, in test_controllers.py in TestAdminController, there are tests for various authentication scenarios.

## Checklist:

- [ ] I have updated the documentation accordingly.
- [x] All new and existing tests passed.

### DIFF
--- a/library_registry/admin/decorators.py
+++ b/library_registry/admin/decorators.py
@@ -1,0 +1,13 @@
+from flask import session
+from functools import wraps
+from library_registry.problem_details import (
+    INVALID_CREDENTIALS,
+)
+
+def check_logged_in(fn):
+    @wraps(fn)
+    def decorated(*args, **kwargs):
+        if not session.get("username"):
+            return INVALID_CREDENTIALS.response # 401 Unauthorized, username or password is incorrect
+        return fn(*args, **kwargs)
+    return decorated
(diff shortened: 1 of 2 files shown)


leonardr (approved):
Take a look at [requires_admin](https://github.com/NYPL-Simplified/circulation/blob/develop/api/admin/routes.py#L59) in circulation to see how we do this in a more complex case.

PR #236, opened 2022-07-06
Add FlaskJWTextended package

nyplwesclark:
This PR adds the Flask-JWT-Extended package to the Pipfile. The Flask-JWT-Extended package will be used to add JWTs and token-based authentication to the library registry.


keithbauer (approved):
I'm ok with this, but I think it'd be useful to see how this will be used. Also, I think this should be pinned to a version.

The other packages should be pinned too, but that is out of scope here and should be handled with a different ticket/branch.

nyplwesclark:
Thanks for approving. Could you explain what you mean by seeing how this will be used? There is a TAD describing the approach for this with a step-by-step outline, and we have several tickets out there describing the work, but we needed to add the package before we could do each ticket. The approach we are using is one task for ticket to make the work easier to review. Happy to add other things to make this understandable if you let me know what they are.

Agreed about the other packages being pinned to specific versions. We attempted that work in [SIMPLY-4093](https://jira.nypl.org/browse/SIMPLY-4093) and it is now backlogged and blocked by [SIMPLY-4092](https://jira.nypl.org/browse/SIMPLY-4092). We should pick it back up and bring it into a sprint after the CM TAD.

Maybe the only place we might disagree is that the authentication package should be pinned to a specific version. One of the many benefits of not writing our own code and instead taking advantage of the eyes and skills of the open-source community is that we can get updates quickly, which is especially important for authentication. If there is a discovered hack, etc, that needs a patch, we want that as soon as possible instead of waiting to manually manage our dependencies. I otherwise agree that pinning packages is a good idea, just not in the case of a topic like authentication or anything depending on cybersecurity.

mwbenowitz:
@nyplwesclark @keithbauer can we reach a compromise position here? What about pegging it so that it won't update any major versions. I agree that allowing it to pick up improvements/patches is a good thing, but a major version bump is likely to/probably will introduce breaking changes and we should be in control of when that happens

keithbauer:
It was just weird seeing only a package added, but after reading the TAD (which I probably should've done sooner) this approach makes sense now.

I'm just a little concerned about breaking changes with version upgrades like we saw with Werkzeug, perhaps we could pin to at least the major version (4.x).

nyplwesclark:
I updated the dependency to include the current major version and any updates to it. That seems like a good compromise.

I agree it seems odd to work on a PR with only one line that adds a package, but we have two engineers about to work on the same repo, same code, same files at the same time, and we need this package to be shared at a minimum so that the two engineers are starting with the same environment. 

Some of the following tickets will also be short but will accomplish a lot. That's because we're using a third-party package where most of the coding has been done, and we're generally configuring instead.

PR #239, opened 2022-07-14
Jwt log in and refresh

thinkwebengineer:
## Description

Implement Flask-JWT-Extended functionality for admin routes.  Modified routes admin/log_in and @check_logged_in decorator have been extended to handle both Flask session and Flask JWT procedures. Add admin/refresh to refresh expiring tokens explicitly. 

## Motivation and Context

[SIMPLY-4132](https://jira.nypl.org/browse/SIMPLY-4132) 
[SIMPLY-4133](https://jira.nypl.org/browse/SIMPLY-4133)
[SIMPLY-4145](https://jira.nypl.org/browse/SIMPLY-4145)
Implement Flask JWT for a more restful API implementing Explicit Header access tokens enabling us to move away from flask sessions. 

## How Has This Been Tested?

Added a tests to confirm functionality of log in and refresh functions and @check_logged_in decorators.

## Checklist:

[x] I have updated the documentation accordingly.
[x] All new and existing tests passed.
[x] Modified code base commented

### DIFF
--- a/library_registry/admin/controller.py
+++ b/library_registry/admin/controller.py
@@ -64,12 +64,13 @@ class AdminController(BaseController):
             return INVALID_CREDENTIALS
 
     def refresh_token(self):
-        if verify_jwt_in_request():
-            identity = get_jwt_identity()
-            access_token = create_access_token(identity=identity)
-            response = Response(201)
-            set_access_cookies(response, access_token)
-            return response
+        if not verify_jwt_in_request(refresh=True):
+            return
+        identity = get_jwt_identity()
+        access_token = create_access_token(identity=identity)
+        response = Response([], 201)
+        set_access_cookies(response, access_token)
+        return response
 
     def log_out(self):
         session["username"] = ""
--- a/library_registry/admin/controller.py
+++ b/library_registry/admin/controller.py
@@ -79,9 +79,6 @@ class AdminController(BaseController):
         if not verify_jwt_in_request(optional=True):
             session["username"] = ""
             return Response(200)
-        response = Response(200)
-        unset_jwt_cookies(response)
-        return response
 
     def refresh_token(self):
         """Refresh JWT access tokens
--- a/library_registry/admin/controller.py
+++ b/library_registry/admin/controller.py
@@ -1,8 +1,8 @@
 from flask import (Response, render_template_string,
-                   session, redirect, request, url_for, make_response)
+                   session, redirect, request, url_for, make_response, jsonify)
 
 from flask_jwt_extended import (create_access_token,
-                                verify_jwt_in_request, get_jwt_identity, set_access_cookies, unset_jwt_cookies)
+                                verify_jwt_in_request, get_jwt_identity, unset_jwt_cookies, create_refresh_token)
 
 from sqlalchemy.orm import (defer, joinedload)
 from library_registry.admin.templates.templates import admin as admin_template
@@ -63,11 +63,11 @@ class AdminController(BaseController):
             return INVALID_CREDENTIALS
         if not jwt_cookie_boolean:
             session["username"] = username
-            return redirect(url_for('admin.admin_view'))
+            return Response(200)
         access_token = create_access_token(identity=username)
-        response = make_response(redirect(url_for('admin.admin_view')), 302)
-        set_access_cookies(response, access_token)
-        return response
+        refresh_token = create_refresh_token(identity=username)
+        return jsonify(access_token=access_token,
+                       refresh_token=refresh_token)
 
     def log_out(self):
         """End point for both Flask Session logout and Flask JWT Logout
--- a/library_registry/admin/controller.py
+++ b/library_registry/admin/controller.py
@@ -40,7 +40,7 @@ class AdminController(BaseController):
         super(AdminController, self).__init__(app)
         self.emailer = emailer_class
 
-    def log_in(self):
+    def log_in(self, jwt_preferred=False):
         """End point for login with requesting flask sessions or JWT tokens
 
         Returns:
@@ -50,10 +50,9 @@ class AdminController(BaseController):
         """
         username = request.form.get("username")
         password = request.form.get("password")
-        jwt = request.form.get('jwt')
         if not Admin.authenticate(self._db, username, password):
             return INVALID_CREDENTIALS
-        if not jwt:
+        if not jwt_preferred:
             session["username"] = username
             return redirect(url_for('admin.admin_view'))
         access_token = create_access_token(identity=username)
--- a/library_registry/admin/controller.py
+++ b/library_registry/admin/controller.py
@@ -1,4 +1,3 @@
-from xml.dom import INVALID_ACCESS_ERR
 from flask import (Response, render_template_string,
                    session, redirect, request, url_for, make_response)
 
@@ -42,47 +41,27 @@ class AdminController(BaseController):
         self.emailer = emailer_class
 
     def log_in(self):
+        """End point for login with requesting flask sessions or JWT tokens
+
+        Returns:
+            err: Invalid credentials
+            flask session: flask session and redirect
+            jwt tokens: JWT tokens as Set-Cookie headers and redirect
+        """
         username = request.form.get("username")
         password = request.form.get("password")
-        if Admin.authenticate(self._db, username, password):
+        jwt = request.form.get('jwt')
+        if not Admin.authenticate(self._db, username, password):
+            return INVALID_CREDENTIALS
+        if not jwt:
             session["username"] = username
             return redirect(url_for('admin.admin_view'))
-        else:
-            return INVALID_CREDENTIALS
-
-    def log_in_with_token(self):
-        """JWT Token based login to create more RESTful API
-
-        Returns:
-            Response: With valid credentials this will return JWT Access and Refresh tokens as Set-Cookie headers and redirects to 'admin/admin_view
-            Response: With invalid credentials this will return the INVALID CREDENTIALS response
-        """
-        username = request.form.get("username")
-        password = request.form.get("password")
-        if Admin.authenticate(self._db, username, password):
-            access_token = create_access_token(identity=username)
-            refresh_token = create_refresh_token(identity=username)
-            response = make_response(
-                redirect(url_for('admin.admin_view')), 302)
-            set_access_cookies(response, access_token)
-            set_refresh_cookies(response, refresh_token)
-            return response
-        else:
-            return INVALID_CREDENTIALS
-
-    def refresh_token(self):
-        """End point for getting new JWT access token if the JWT Refresh token is still valid
-
-        Returns:
-            Response: With a valid JWT Refresh token, this will return a new JWT access token as Set-Cookie header
-            Response: With invalid Refresh or no token this will return and INVALID_ACCESS_ERR
-        """
-        if not verify_jwt_in_request(refresh=True, optional=True):
-            return INVALID_CREDENTIALS
-        identity = get_jwt_identity()
-        access_token = create_access_token(identity=identity)
-        response = Response([], 201)
+        access_token = create_access_token(identity=username)
+        refresh_token = create_refresh_token(identity=username)
+        response = make_response(
+            redirect(url_for('admin.admin_view')), 302)
         set_access_cookies(response, access_token)
+        set_refresh_cookies(response, refresh_token)
         return response
 
     def log_out(self):
--- a/library_registry/admin/controller.py
+++ b/library_registry/admin/controller.py
@@ -1,8 +1,8 @@
 from flask import (Response, render_template_string,
-                   session, redirect, request, url_for, make_response, jsonify)
+                   session, request, jsonify)
 
 from flask_jwt_extended import (create_access_token,
-                                verify_jwt_in_request, get_jwt_identity, unset_jwt_cookies, create_refresh_token)
+                                verify_jwt_in_request, get_jwt_identity, create_refresh_token)
 
 from sqlalchemy.orm import (defer, joinedload)
 from library_registry.admin.templates.templates import admin as admin_template
--- a/library_registry/admin/controller.py
+++ b/library_registry/admin/controller.py
@@ -83,6 +83,13 @@ class AdminController(BaseController):
         unset_jwt_cookies(response)
         return response
 
+    def refresh(self):
+        if not verify_jwt_in_request(optional=True):
+            return INVALID_CREDENTIALS
+        identity = get_jwt_identity()
+        access_token = create_access_token(identity=identity)
+        return jsonify(access_token=access_token)
+
     def libraries(self, live=True):
         # Return a specific set of information about all libraries in production;
         # this generates the library list in the admin interface.
--- a/library_registry/admin/controller.py
+++ b/library_registry/admin/controller.py
@@ -1,5 +1,5 @@
 from flask import (Response, render_template_string,
-                   session, request, jsonify)
+                   session, request, jsonify, redirect, url_for)
 
 from flask_jwt_extended import (create_access_token,
                                 verify_jwt_in_request, get_jwt_identity, create_refresh_token)
@@ -63,7 +63,7 @@ class AdminController(BaseController):
             return INVALID_CREDENTIALS
         if not jwt_cookie_boolean:
             session["username"] = username
-            return Response(200)
+            return redirect(url_for('admin.admin_view'))
         access_token = create_access_token(identity=username)
         refresh_token = create_refresh_token(identity=username)
         return jsonify(access_token=access_token,
@@ -78,7 +78,7 @@ class AdminController(BaseController):
         """
         if not verify_jwt_in_request(optional=True):
             session["username"] = ""
-            return Response(200)
+            return redirect(url_for('admin.admin_view'))
 
     def refresh_token(self):
         """Refresh JWT access tokens
--- a/library_registry/admin/controller.py
+++ b/library_registry/admin/controller.py
@@ -83,8 +83,8 @@ class AdminController(BaseController):
         unset_jwt_cookies(response)
         return response
 
-    def refresh(self):
-        if not verify_jwt_in_request(optional=True):
+    def refresh_token(self):
+        if not verify_jwt_in_request(optional=True, refresh=True):
             return INVALID_CREDENTIALS
         identity = get_jwt_identity()
         access_token = create_access_token(identity=identity)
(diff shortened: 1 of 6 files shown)


library_registry/admin/controller.py
```diff
@@ -42,9 +50,54 @@ def log_in(self):
...
+            refresh_token = create_refresh_token(identity=username)
+            response = make_response(
+                redirect(url_for('admin.admin_view')), 302)
+            set_access_cookies(response, access_token)
+            set_refresh_cookies(response, refresh_token)
+            return response
+        else:
+            return INVALID_CREDENTIALS
+
+    def refresh_token(self):
```
nyplwesclark: If our clients use cookies, they are going to expect cookie-like behavior. This refresh setup requires setting up some checks on the frontend to check if an access token is expired and then redirect to the refresh endpoint to get the new access token. Normal cookie behavior would not require this redirect and instead would update the Expire attribute, so I'm not sure how a client using cookies would make it to the refresh endpoint without also including unexpected logic on the frontend. Have you looked at [implicit refresh with cookies](https://flask-jwt-extended.readthedocs.io/en/stable/refreshing_tokens/#implicit-refreshing-with-cookies)? To support cookies, let's modify this and surrounding code to mimic expected cookie behavior and then retest.

library_registry/admin/routes.py
```diff
@@ -10,59 +13,85 @@
...
 @admin.route('/admin/log_out')
 @check_logged_in
 @returns_problem_detail
+@jwt_required(optional=True)
 def log_out():
     return current_app.library_registry.admin_controller.log_out()
 
+
 @admin.route('/admin/libraries')
 @check_logged_in
```
nyplwesclark, line 122: Any admin route that has the check_logged_in decorator should probably also have the jwt_required decorator.

mwbenowitz: Is it possible to roll the functionality for both `check_logged_in` and `jwt_required` into a single method? Depending on how a user logs in one will succeed and one will fail right now

library_registry/admin/controller.py
```diff
@@ -42,9 +50,54 @@ def log_in(self):
...
+        username = request.form.get("username")
+        password = request.form.get("password")
+        if Admin.authenticate(self._db, username, password):
+            access_token = create_access_token(identity=username)
+            refresh_token = create_refresh_token(identity=username)
+            response = make_response(
+                redirect(url_for('admin.admin_view')), 302)
+            set_access_cookies(response, access_token)
+            set_refresh_cookies(response, refresh_token)
+            return response
```
mwbenowitz: We should consider how we need this returned (I think some conversations about this have already started). The log in/out/refresh methods should be able to interact with web, mobile and API clients, and to me the best way to do that would be to return the two tokens explicitly. But I'm sure there are factors I'm not considering!

library_registry/admin/controller.py
```diff
@@ -1,4 +1,10 @@
-from flask import (Response, render_template_string, session, redirect, request, url_for)
+from xml.dom import INVALID_ACCESS_ERR
```
mwbenowitz: I don't think this is used? It's referenced in a docstring, but it seems to be the wrong error class. (`INVALID_CREDENTIALS` is returned which seems right to me)

library_registry/admin/routes.py
```diff
@@ -10,59 +13,85 @@
...
 @admin.route('/admin/log_in', methods=["POST"])
 @returns_problem_detail
 def log_in():
     return current_app.library_registry.admin_controller.log_in()
 
+
+@admin.route('/admin/log_in_w_token', methods=['POST'])
+@returns_problem_detail
+def log_in_w_token():
+    return current_app.library_registry.admin_controller.log_in_w_token()
```
mwbenowitz: Would it be a good idea to unify this with the original `/log_in` method? Hypothetically we could set a session variable _and_ create tokens and therefore allow interaction using either mechanism?

Or maybe this could accept a parameter that specifies the type of authentication we're doing? That way we could let applications choose but not require the additional endpoint.

Can see arguments for all sides!

mwbenowitz (commented):
Some thoughts about the approach, mainly about integrating the two authentication methods and how we want to return tokens, but overall this looks really great! Appreciate attention to detail on the docstrings, tests and everything else.

library_registry/admin/routes.py
nyplwesclark: This is a good point. Keeping the session object and related right now will need these workarounds. These could become easy to forget so we'll need to remember to clean them up in the future.

library_registry/admin/controller.py
thinkwebengineer: Correct, removed

library_registry/admin/routes.py
thinkwebengineer: Great idea, I have extended the log_in function to check for a 'jwt': boolean in the form. And handle both routes.

nyplwesclark: Hey Bob. Your approach is a good first attempt at this. The one downside is that it assumes the client is going to have a toggle or box to check if they want jwt, and that isn't really something we can expect. Have you tried the parameter approach? The client can pass in how they want to authenticate to the `/admin/log_in` route, and you'll just need to pass that to your `log_in()` method in the controller and handle there (it looks like you already have the shape of it all in place there...great!). Bonus: if we need to add more authentication options, we can make that happen easily with this approach.

thinkwebengineer: Great idea. I have updated accordingly.

library_registry/admin/routes.py
thinkwebengineer: Great idea

library_registry/admin/controller.py
thinkwebengineer: Implemented.

library_registry/admin/routes.py
```diff
@@ -10,59 +14,91 @@
...
+        return response
+
+
 @admin.route('/admin/', strict_slashes=False)
 def admin_view():
     return current_app.library_registry.view_controller()
 
+
 @admin.route('/admin/log_in', methods=["POST"])
+@admin.route('/admin/log_in/<jwt_preferred>')
```
mwbenowitz: Should we consider making this a URL parameter so that other devs clearly understand what they're requesting?

library_registry/admin/routes.py
```diff
@@ -10,59 +14,91 @@
...
-def log_in():
-    return current_app.library_registry.admin_controller.log_in()
+def log_in(jwt_preferred=False):
+    return current_app.library_registry.admin_controller.log_in(jwt_preferred)
+
 
 @admin.route('/admin/log_out')
 @check_logged_in
 @returns_problem_detail
+@jwt_required(optional=True)
```
mwbenowitz: With the new combined approach above this probably isn't required, right?

library_registry/admin/controller.py
```diff
@@ -18,33 +23,66 @@
...
+            return INVALID_CREDENTIALS
+        if not jwt_preferred:
             session["username"] = username
             return redirect(url_for('admin.admin_view'))
-        else:
-            return INVALID_CREDENTIALS
+        access_token = create_access_token(identity=username)
+        response = make_response(
+            redirect(url_for('admin.admin_view')), 302)
+        set_access_cookies(response, access_token)
```
mwbenowitz: wait, I'm afraid I've led us down a bad path here. When we de-couple the admin interface and deploy it as a standalone app this redirect won't function anymore, correct? We _would_ need another endpoint to directly return the token in that case if we need to support redirects and an explicit token return flow.

library_registry/admin/routes.py
```diff
@@ -10,59 +14,91 @@
     'admin', __name__,
     template_folder='templates')
 
+# Using an `after_request` callback, we refresh any token that is within 30
+# minutes of expiring. Change the timedeltas to match the needs of your application.
+
+
```
mwbenowitz: nit: extra lines here

library_registry/admin/routes.py
```diff
@@ -10,59 +14,91 @@
...
+@admin.after_request
+def refresh_expiring_jwts(response):
+    try:
+        exp_timestamp = get_jwt()["exp"]
+        now = datetime.now(timezone.utc)
+        target_timestamp = datetime.timestamp(now + timedelta(minutes=30))
```
mwbenowitz: how long is the default lifespan for a token? I think it's ~10 hours? If it's shorter than that we should make this correspondingly shorter (imo).

tests/test_decorators.py
```diff
@@ -82,6 +97,27 @@ def returns_already_encoded_testview():
...
+    def refresh_expiring_jwts(response):
+        try:
+            exp_timestamp = get_jwt()["exp"]
+            now = datetime.now(timezone.utc)
+            target_timestamp = datetime.timestamp(now + timedelta(minutes=30))
+            if target_timestamp > exp_timestamp:
+                access_token = create_access_token(identity=get_jwt_identity())
+                response = make_response(response, 201)
+                set_access_cookies(response, access_token)
+            return response
```
mwbenowitz: We should import the method from the actual implementation and use it here. feels very easy to forget to update this if we change the function.

mwbenowitz (commented):
This is looking really good! Really just one big question about how we handle authentication after "decoupling" the front and back ends, otherwise just some minor things.

library_registry/admin/routes.py
nyplwesclark: The default lifespan provided by the package we're using for access tokens is 15 minutes. When we were talking about headers and the explicit refresh process, we discussed that the access token was 1 hour and the refresh token was 30 days. Now that we're supporting cookies and the implicit refresh process, we've had conversations about the best expiration time but haven't nailed down an exact one. It seems like we're somewhere between 8 and 12 hours and should be easy enough to update that when we have it.

library_registry/admin/controller.py
nyplwesclark: I'm not sure I'm following so let me know if I haven't addressed what you mean. We are using `redirect` here for the Response object it returns to help build a larger response in `make_response`. I'm not sure if `return redirect(url_for('admin.admin_view'))` would stop working since it just returns a response. It certainly wouldn't do what was expected for login because there'd be no cookie sent to the client if that's what you mean. That's why we're explicitly building the larger response instead with the 302 and the cookie.

I don't think we need to support redirects, but we should have that conversation with frontend to make sure they are prepared to handle them. It seems there shouldn't be too much of a change from how they read 302s now. You are right that we will need another `/refresh` endpoint when we implement the explicit token return flow (the current cookies implementation uses the implicit refresh process to behave in a more expected way), and the logic for hitting that refresh endpoint has to be implemented on the frontend.

tests/test_decorators.py
nyplwesclark: How very DRY 😆😆😆 +1

library_registry/admin/routes.py
nyplwesclark: Hey Mike...could you clarify here? Do you mean that it's difficult to tell that a Boolean should be passed here? Should we be setting defaults on the routes as optional parameters? Is there a different syntax you had in mind?

Or do you mean not Flask URL parameters but general query URL parameters?

library_registry/admin/routes.py
thinkwebengineer: Missed the duplication, thanks.

library_registry/admin/controller.py
thinkwebengineer: If we remove redirects from our application flow, then I think the best option that can plug and play with both library_reg and CM is Explicit cookies.  However, with the library_reg I think that we are only having to support web browsers for the protected admin route.  If this is the case, and we aren't planning a massive expansion of this system, implicit cookies work for them very automagically.  Even if we decide to pull redirects, we just need to pass a Response object with set_cookie called on it. 

Edit: @nyplwesclark has pointed out to me that Explicit and Implicit are not mutually exclusive as I think my brain was trying to make them.  I think this might clear up some confusion on this point.

library_registry/admin/routes.py
thinkwebengineer: We can set up our preferred JWT expiration times in the application factory set-up in app.py.  This function is checking access_cookies expiring in 30 minutes and sends a new one.  We still need clarification on how long the access cookies need to exist for and if we want to change the refresh route to update with a smaller or wider window.

tests/test_decorators.py
thinkwebengineer: Great idea, thanks for the clarification @nyplwesclark . Commited.

library_registry/admin/controller.py
thinkwebengineer: From what I understand of this, you are correct, this will require a bit of changes that would affect the front end too however.  We would need to remove redirects to explicitly send access and refresh tokens. And front ends relying on our current flow would be affected accordingly.

library_registry/admin/controller.py
nyplwesclark: The backend cannot handle redirects directly, but the backend is still in charge of the response status codes and any relevant URLS that need to be returned. You can't eliminate sending proper response codes. I think there is some confusion here around what it means to handle redirects vs the forming of responses, and that confusion may be mine. Our API can't handle redirects and absolutely shouldn't even try. It absolutely should return 302s and URLs when relevant.

By explicit cookies, do you mean explicit refresh? Explicit refresh is the recommended approach for supporting headers, which is originally the only thing we were going to support. Implicit refresh is the recommended approach for cookies, which is what we then decided to add support for and focus on first. Explicit refresh requires refresh tokens, and the documentation clearly states that explicit refresh tokens are not recommended for cookies. If you look at the setup of the code for explicit refresh for cookies, it becomes clear in my opinion why it's a bad design choice....it doesn't behave like cookies at all. It seems we keep coming back to the idea of explicit refresh with cookies, and I apologize for the changing requirements that may have confused that, but I think we should continue with the approaches that have already been outlined for cookies.

library_registry/admin/controller.py
mwbenowitz: yeah, that's the point I'm stuck on, I haven't been able to land on a satisfying solution that supports both front-ends. Let's discuss and see if we can figure something out.

library_registry/admin/routes.py
mwbenowitz: The general URL params so `/log_in?jwt_preferred=true` Which is less explicit in the code because it has to be retrieved from the `request.args` object but is more explicit for people sending requests to this endpoint.

My concern is that `log_in/false` doesn't indicate what that boolean is doing to outside users/applications.

library_registry/admin/routes.py
mwbenowitz: Let's leave it as is for now. I don't think it's a huge issue and it seems like it will be easy to adjust in the future!

tests/test_decorators.py
mwbenowitz: the `refresh_expiring_jwts` method

library_registry/admin/controller.py
thinkwebengineer: I think the satisfying solution is to combine both.  Please check how I have updated the `@after_request` decorator to specify for cookies.  I can do the same check for the explicit refresh end point as well to prevent cookie users from requesting explicitly.

library_registry/admin/routes.py
thinkwebengineer: Updated to <jwt_cookie_boolean> to be more clear as to what the end point is expecting.

library_registry/admin/controller.py
thinkwebengineer: At the moment we will leave our current redirects, as this does not need to be changed to implement the extension.  With the explicit tokens the log_in will be expanded and only return a token.  Leaving redirect up to the front end.

library_registry/admin/controller.py
```diff
@@ -18,33 +23,65 @@
...
-        if Admin.authenticate(self._db, username, password):
+        if not Admin.authenticate(self._db, username, password):
+            return INVALID_CREDENTIALS
+        if not jwt_cookie_boolean:
             session["username"] = username
             return redirect(url_for('admin.admin_view'))
-        else:
-            return INVALID_CREDENTIALS
+        access_token = create_access_token(identity=username)
+        response = make_response(redirect(url_for('admin.admin_view')), 302)
```
mwbenowitz: still have to raise a question here! when requesting a JWT I still don't think the redirect makes sense. This will redirect to the `/admin` route that serves the locally deployed version of the admin interface. If the request _doesn't_ come from that instance, it breaks right?

I think the right thing to do would be to return a `200` with the token in the body of the response.

nyplwesclark: I admit that it took me a bit and probably too long to wrap my head around this because I was thinking in terms of monolithic software vs a decoupled architecture. I think the example that made me finally understand why **this is the right approach** is: we're building and maintaining an API. In theory, multiple clients could be calling our API. It makes no sense to send back to everyone a redirect to a URL that is specific to only one particular client. Thanks very much, Mike!!!!

thinkwebengineer: Ok, so originally we wanted Headers, but then I didn't ask the question of if we wanted to keep the redirects.  This brought in the solution of cookies, to retain functionality that we don't want.  With this functionality gone, headers becomes a very easy thing to implement and is the recommended use of this package for API only use case.  Please review the newly updated functionality of explicit token headers and refresh.  I still need to add end point documentation before this is finished. But I think this is what was originally intended.

library_registry/admin/routes.py
```diff
@@ -10,59 +13,156 @@
...
 
 @admin.route('/admin/log_out')
 @check_logged_in
 @returns_problem_detail
 def log_out():
     return current_app.library_registry.admin_controller.log_out()
 
+
 @admin.route('/admin/libraries')
+@jwt_required(optional=True)
```
mwbenowitz: nit: line still dangling here!

tests/test_decorators.py
```diff
@@ -13,18 +19,28 @@
...
 
 @pytest.fixture
 def app():
-    return Flask(__name__)
+    app = Flask(__name__)
+    babel = Babel()
+    babel.init_app(app)
+    app.config['JWT_SECRET_KEY'] = 'testing'
+    jwt = JWTManager(app)
+    app.config['JWT_TOKEN_LOCATION'] = ['cookies', 'headers']
```
mwbenowitz: for consistencies sake should remove `cookies` here

mwbenowitz (approved):
👍 This looks good to me! Two housekeeping things but otherwise good to go. Especially appreciate the doc strings.

nyplwesclark (approved):

PR #258, opened 2023-01-05
Lock in dependencies for stability

thinkwebengineer:
This locks in our Pipfile to specific versions of the dependencies we require to help maintain stability of the project.


Pipfile
```diff
@@ -4,36 +4,36 @@ verify_ssl = true
...
 SQLAlchemy = "==1.3.23"
-Pillow = "*"
-pytest = "*"
-pytest-datadir = "*"
-bcrypt = "*"
-gunicorn = "*"
+Pillow = "==9.4.0"
+pytest = "==7.2.0"
+pytest-datadir = "==1.4.1"
+bcrypt = "==4.0.1"
```
mwbenowitz, line 27: I'm worried about this one, it seems to _maybe_ cause some issues. Would there be an issue locking it to 3.x?

mwbenowitz (approved):

PR #262, opened 2024-07-31
TGR-53 Update OS

keithbauer:
* Python 3.9 to Python 3.12
* Alpine 3.13 to Alpine 3.20
* Nginx 1.19.8 to 1.27.0
* NJS 0.5.2 to 0.8.5
* Update lxml, feedparser, and psycopg2-binary
* Remove npm / registry 1.4.x assets

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -153,22 +153,6 @@ RUN set -ex \
  && pipenv install --dev --skip-lock --clear \
  && apk del --no-network .build-deps
 
-# Build a static version of the front end to serve
-COPY ./package*.json ./
-
-RUN set -ex \
- && apk add --no-cache --virtual .node-build-deps \
-    make \
-    build-base \
-    npm \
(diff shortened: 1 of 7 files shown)


Dockerfile
```diff
@@ -126,7 +126,7 @@ COPY ./Pipfile* ./
 #
 #   https://github.com/pypa/pipenv/issues/1226#issuecomment-598487793
 #
-ENV WORKON_HOME /simplified_venv
+ENV WORKON_HOME=/simplified_venv
```
EdwinGuzman, line 129: Does this update mean this wasn't working before or it's just a different syntax?

EdwinGuzman (approved):
Just one question but not a blocker.

Dockerfile
keithbauer: It's only updated syntax, it should work as it did previously.

